PR #20, opened 2016-11-03
build(webpack): @angular/layouts webpack alias

joshwiens:
- Custom resolver to allow library-like import syntax as opposed to relative file path

screen shot 2016-11-03 at 2 22 31 am


d3viant0ne:
From a webpack standpoint it works when you bundle and run the demo. That said, something like `vscode` is always going to read that alias as an error as the alias isn't valid until you fire up the dev server and bundle everything. For a smarter IDE like Webstore, it's more annoying than a little red squiggly. 

I closed the PR because given the side effects for the `nice to have` I was pretty sure you weren't going to actually want it in #master

PR #31, opened 2016-11-28
chore(config): update development configs

d3viant0ne:
* Ignores `vscode` in gitignore
* Adds an editorconfig with the https://github.com/angular/angular settings

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+# http://editorconfig.org
+
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+insert_final_newline = false
+trim_trailing_whitespace = false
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@d3viant0ne - If the `.vscode/` a directory of editor configurations, it should not be in github; so the ignore is correct. Why do you want to version manage the .vscode ?

d3viant0ne:
I set it to ignore the `.vscode` directory. As it stands right now, if you use the bundled TypeScript version with current Visual Studio Code or create launch / debug configurations ...

screen shot 2016-11-28 at 1 29 10 pm

You end up with ...

screen shot 2016-11-28 at 1 29 29 pm

It should be ignored unless you want to add a vscode configuration for the project.

As far as the editor configuration isn't specific to vscode, and just simplifies basic code formatting for projects. The configuration that is included is a `cmd + c` of that which is currently in angular.

d3viant0ne:
I can certainly add a `vscode` launch / debug configuration if you prefer that to ignoring it and letting people use their own.

Issue #39, opened 2016-12-12
 npm ERR! 404  '@angular/flex-layout' is not in the npm registry. npm ERR! 404 You should bug the author to publish it (or use the name yourself!)


smile2014:
npm ERR! 404  '@angular/flex-layout' is not in the npm registry.
npm ERR! 404 You should bug the author to publish it (or use the name yourself!)

fxck:
I wonder why..

https://github.com/angular/flex-layout/issues/35

https://github.com/angular/flex-layout/issues/30

![image](https://cloud.githubusercontent.com/assets/1303561/21097777/856421c2-c066-11e6-9469-6cb4cdda523b.png)

kuncevic:
I know a few more people including me who just cath that. It would of be great if the message **peanding feature** would of be better highlighted.

ThomasBurleson:
Fixed with SHA d3ea8ad.

smile2014:
thank you

2016-12-12 21:19 GMT+08:00 Thomas Burleson <notifications@github.com>:

Issue #35, opened 2016-12-09
install fails!


lgt:
so I try to install over npm 

`npm install @angular/flex-layout -save`

but I get 

`'@angular/flex-layout' is not in the npm registry.`

aveniri:
read the docs, it hasn't been published on npm yet

lgt:
actually I took from docs `npm install @angular/flex-layout -save`

https://github.com/angular/flex-layout#fast-start

lgt:
sorry I see now 
pending future

Issue #30, opened 2016-11-28
(bug): Package not published on npm


amcdnl:
The package is not published on npm

DzmitryShylovich:
https://github.com/angular/flex-layout#fast-start
> Note: With gulp build:components the files are depolyed to node_modules/@angular/flex-layout. The formal deploy process to npm (and use of npm i @angular/flex-layout) is not yet available and is pending public release of the Github repository!

d3viant0ne:
@DzmitryShylovich - It's not actually on npm per that documentation. Given it's pre-alpha state, i'm betting that's on purpose.

@amcdnl - I'm running a local build of it on an internal prod app where a pre-alpha oops isn't a big deal. What it does it does surprisingly well given that whole `pre-alpha` thing.

trainerbill:
To get this working I did the following:

1.  Clone the repo
2.  Remove "noEmitHelpers": true from tsconfig.  May be a bug?
3.  gulp build:components
4.  change package.json main and add types
```
"main": "dist/@angular/flex-layout/index.js",
"types": "dist/@angular/flex-layout/index.d.ts",
```
5.  copy the directory to your ng2 app node_modules folder
6.  Continue:  https://github.com/angular/flex-layout#application-usages

anyone know when this will hit NPM?

aberwag:
You can juste add this repo to your dependencies 

run this command (yout can change #master to your favorite branch or tag :wink:  )
```
npm install --save https://github.com/angular/flex-layout.git#master
```
result:
```
...
"@angular/flex-layout": "git+https://github.com/angular/flex-layout.git#master",
...
```

ThomasBurleson:
This will be resolve when v1.0.0-beta.1 is published to NPM.

trainerbill:
@ThomasBurleson Is there a timeframe on when a first version will be published to npm?  days/weeks/months?

amcdnl:
@aberwag - side note: u have to build it, that install doesn't run a build.

ThomasBurleson:
Use `gulp build:lib` and find the deployed files in `./dist/@angular/flex-layout`

ThomasBurleson:
Note: earlier commands have been deprecated. Please use `gulp build:release`

Issue #16, opened 2016-11-03
namespace prefix - identify best prefix use in markup and code


ThomasBurleson:
The current directives use `fx-` prefixes with as HTML markup:

*  `fx-layout`
*  `fx-layout-wrap`
*  `fx-layout-align`
*  `fx-flex`             
*  `fx-flex-fill`
*  `fx-flex-order`
*  `fx-flex-offset`
*  `fx-flex-align`

#### Questions

* Is the `fx-` namespace the best choice ?
* What about the `fxOnMediaQueryChanges` lifecycle method ?

ryancampbell:
Honestly I'd prefer to stick with flex- instead of fx-

ThomasBurleson:
Will not change. 

> Just as an aside this would be things like `flex-flex`... not good.

ThomasBurleson:
Angular team decision to use the `fx-` namespace prefixes.

ryancampbell:
Obviously you would just do flex-align, not flex-flex-align. Just like it was in Material 1. There was no good reason to change it, but sure fx- it is.

PR #40, opened 2016-12-12
chore(readme): put info about npm status on top

fxck:
should prevent these 

https://github.com/angular/flex-layout/issues/39
https://github.com/angular/flex-layout/issues/35
https://github.com/angular/flex-layout/issues/30

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,8 @@
+# 👷 not yet released on NPM
+See this [comment](https://github.com/angular/flex-layout/issues/30#issuecomment-265864269) for directions about building flex-layout manually.
+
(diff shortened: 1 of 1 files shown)


thmb:
Will beta version keep "fx" prefix for flex layout components or will it be change to "md" or something else?
What is the idea behind using another prefix if so? Thanks in advance.

fxck:
@thmb this is separate from material design, so definitely not `md`, as for fx vs flex, there was an issue about that https://github.com/angular/flex-layout/issues/16 and the result is that it will stay `fx`

PR #56, opened 2016-12-21
chore: es5 modules as entry point

DevVersion:
* Fixes inconsistent release packaging between `@angular/{core,material}`


ThomasBurleson:
@DevVersion - lgtm. thx.

PR #61, opened 2016-12-23
Consistent naming / terms usage

webmaxru:
Set of terms proposed/updated:
- Angular Flex Layout (or just Flex Layout) - used in general context. Neither Angular Flexbox Layout nor Flex-Layout
- Flex Layout API - used in more technical context, when we use code samples
- Flexbox (not FlexBox)

It would be nice to update the Wiki page accordingly


ThomasBurleson:
@webmaxru - lgtm. Thank you sir.

PR #94, opened 2017-01-05
chore(demo-app): remove hammer.js from assets

d3viant0ne:
Removes the minified version of hammer.js in the assets folder of the demo-app as it is already using a CDN.


jelbourn:
LGTM

PR #98, opened 2017-01-06
chore: only publish artifacts once

DevVersion:
Right now Travis tries to publish the build artifacts to the `flex-layout-builds` repository for each mode.
Fortunately it still only published once, because Git doesn't allow re-publishing of the same git tags.

The issue has been fixed with version `0.0.7` of `travis-after-modes` (DevVersion/travis-after-modes@387328e)

This is similar to angular/material2@8470534


ThomasBurleson:
@DevVersion - thank you.

PR #99, opened 2017-01-06
Fix rxjs import

bowenni:


jelbourn:
LGTM

FYI @ThomasBurleson, importing from `'rxjs'` alone brings in the entire library with all operators vs. justing bringing in the parts you're using. We have a check in the material2 CI that blocks doing this.

PR #111, opened 2017-01-13
build: cleanup travis ci setup

DevVersion:
* Properly reports exit codes of the unit tests to Travis
* Updates Karma version
* Uses Yarn Installation from Travis CI container
* Runs Unit tests now against the latest Chrome & Firefox versions.
* Made the required tests less, so the tests finish quicker and run more stable.


DevVersion:
@ThomasBurleson Okay, as discussed, the CI is now exiting properly and the cleanup should be ready. And the failing unit tests are still present ;)

PR #121, opened 2017-01-21
fix(fxHide,fxShow): fix standalone breakpoint selectors

### LINKED ISSUE #59: template parse error when using only responsive selectors.
Developers currently must use the default API when one or more responsive APIs are defined.

Consider the following scenario where the default is `fxShow` and the responsive API defined is `fxShow.xs`:

```html
<div fxShow [fxShow.xs]="menuOpen" class="banner">
  I should hide and show
</div>  
```

The above ^ works fine. When I do not, however, specify the default non-responsive selector `fxShow`:

```html
<div [fxShow.xs]="menuOpen" class="banner">
  I should hide and show
</div>  
```

 then a Template Parse error occurs:

```terminal
Uncaught (in promise): Error: Template parse errors:
Can't bind to 'fxShow.xs' since it isn't a known property of 'div'. ("
    </button>
    
    <div [ERROR ->][fxShow.xs]="menuOpen" class="banner">
      I should hide and show
    </div>  
"): TestApp@5:9
Error: Template parse errors:
Can't bind to 'fxShow.xs' since it isn't a known property of 'div'. ("
    </button>
```

Here is the [Plunkr Demo](https://plnkr.co/edit/pZHpB08iaLiO865ieWPI?p=preview) that shows this.

> NOTE: this has also been reported in [/angular/issues/13645](https://github.com/angular/angular/issues/13645)

### LINKED ISSUE #62: selector fxHide.xs doesn't work
fxHide.[any-size] doesn't work (It doesn't hide an element)

### LINKED ISSUE #105: The hover circle on material md-icon button is off when using fx-show / hide
When using fx-show / hide in a md-icon-button, the hover circle is a few pixels off. Here's a screenshot:
image

ThomasBurleson:
* Many directive selectors were missing a `,` separator between the breakpoint-enabled selectors. 
  >  xs and gt-xs selectors were not delimited.
* **fxShow**, **fxHide** preserve and restore the element's original display CSS style
* added more tests standalone breakpoint selectors, disabled selectors, and deactivated mediaQueries
* The **fxHide** and **fxShow** should not change the display mode to `flex`. Instead
the display mode (when not `none`) should be the default (=== `block`) or the explicitly assigned display mode for that element.

fixes #62, fixes #105, closes #59.

### DIFF
--- a/src/demo-app/app/docs-layout-responsive/responsiveRowColumns.demo.ts
+++ b/src/demo-app/app/docs-layout-responsive/responsiveRowColumns.demo.ts
@@ -25,8 +25,7 @@ import {MatchMediaObservable} from "../../../lib/media-query/match-media";
                     [fxLayout.gt-lg]="firstColGtLg" 
                     fxFlex="50%" 
                     fxFlex.gt-sm="25"
-                    fxShow="true"
-                    fxShow.md="false" 
+                    fxHide.md 
                     (click)="toggleLayoutFor(1)" style="cursor: pointer;">
                 <div fxFlex>Col #1: First item in row</div>
                 <div fxFlex>Col #1: Second item in row</div>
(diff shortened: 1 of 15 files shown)


ThomasBurleson:
@jelbourn - ping for the caretaker.

PR #143, opened 2017-01-30
fix(api): restore element's display mode and more...

### LINKED ISSUE #140: fxHide / fxShow ignores breakpoint aliases
Tried in both FF & Chrome.
The following span is always hidden, although it should be shown when 'screen and (min-width: 960px) and (max-width: 1279px)'.

    <span fxHide fxShow.gt-xs>Text</span>

Another example:

    		<span fxHide.lg>text</span>

Never hides.

### LINKED ISSUE #141: fxHide breaks style of non-"display: box" elements
As part of a responsive UI, there are times when we want to hide some elements of our UI (not whole parts, just a bunch of elements e.g. an item's label) on small screens.

As an example, I'd expect the following to work so that on non-xs screens, it displays the label inline next to the icon, and on xs screens only the icon remains:

    <div>
        
        <span fxHide.xs>Label</span>
    </div>

The show/hide functionality now works as of `v2.0.0-beta.4`, but it looks like it toggles only between `display: none` and `display: block`, ignoring the actual display mode the element was in before `fxHide` got applied.

I think `fxHide` should take into account the element's actual display mode and not interfere with it.

ThomasBurleson:
*  fxShow and fxHide should restore original display modes when deactivating
*  combination uses of fxHide + fxShow should work properly
*  static uses of `fxHide=""` should hide the hosting elmement

fixes #140, fixes #141.

### DIFF
--- a/src/lib/flexbox/api/hide.spec.ts
+++ b/src/lib/flexbox/api/hide.spec.ts
@@ -17,7 +17,7 @@ import {ObservableMediaService} from '../../media-query/observable-media-service
 import {BreakPointsProvider} from '../../media-query/providers/break-points-provider';
 import {BreakPointRegistry} from '../../media-query/breakpoints/break-point-registry';
 
-import {customMatchers, expect} from '../../utils/testing/custom-matchers';
+import {customMatchers, expect, NgMatchers} from '../../utils/testing/custom-matchers';
 import {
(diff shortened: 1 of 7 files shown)


ThomasBurleson:
@d3viant0ne - please review and comment with LGTM if all good. Thx.

PR #145, opened 2017-01-30
fix(fxLayoutGap): skip hidden element nodes

### LINKED ISSUE #136: fxLayoutGap should consider hidden elements
When an item in a layout is hidden (usually conditionally) then fxLayoutGap should not apply a gap to the first displayed element

```html
<style>
   .col1{
    display:none !important;
  }
</style>
<div class="container" fxLayout="row" fxLayoutGap="16px">
  <div fxFlex class="col1">Div 1</div>
  <div fxFlex class="col2">Div 2</div>
  <div fxFlex class="col3">Div 3</div>
</div>
```

In this example `Div 2` should not have a gap in front of it.

ThomasBurleson:
fxLayoutGap should consider (and skip) hidden elements when applying layout-gap logic.

fixes #136.

### DIFF
--- a/src/lib/flexbox/api/base.ts
+++ b/src/lib/flexbox/api/base.ts
@@ -6,6 +6,10 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import {ElementRef, Renderer, OnDestroy} from '@angular/core';
+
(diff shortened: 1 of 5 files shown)


ThomasBurleson:
@d3viant0ne - please review.

PR #148, opened 2017-01-31
fix(fxLayoutGap): consider fxLayoutWrap to apply gap logic

### LINKED ISSUE #108: fxLayout="row" fxLayoutWrap fxLayoutGap="10px" gives an offset starting at the 2nd line
```
<div fxLayout="row" fxLayoutWrap fxLayoutGap="10px">
  <div class="card" fxFlex="25%" *ngFor="let i of [1, 2, 3, 4, 5, 6, 7, 8, 9]">
    Test
  </div>
</div>
```
image

https://plnkr.co/edit/f44sdpW159qKSGvWkWm5?p=preview

(probably) related to #106 except that here the layout is set to `row`

ThomasBurleson:
*  `margin-right` be used with `fxLayout="row" fxLayoutWrap fxLayoutGap` combinations
*  `margin-bottom` be used with `fxLayout="column" fxLayoutWrap fxLayoutGap` combinations

fixes #108

### DIFF
--- a/src/lib/flexbox/api/layout-gap.spec.ts
+++ b/src/lib/flexbox/api/layout-gap.spec.ts
@@ -7,7 +7,7 @@
  */
 import {Component, OnInit} from '@angular/core';
 import {CommonModule} from '@angular/common';
-import {TestBed, ComponentFixture} from '@angular/core/testing';
+import {TestBed, ComponentFixture, async} from '@angular/core/testing';
 
 import {BreakPointsProvider} from '../../media-query/breakpoints/break-points';
 import {BreakPointRegistry} from '../../media-query/breakpoints/break-point-registry';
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@here - merge **after** PR 145.

ThomasBurleson:
@d3viant0ne - please review.

maxime1992:
@ThomasBurleson I tried your PR on my demo and it's great :smile: 

One thing tho, when elements are going to the next line (fxLayoutWrap), they don't have a margin-top.

I think it would be a good idea to apply the gap not only horizontally between elements but also vertically, don't you think ?

ThomasBurleson:
@maxime1992 - the gap will only be applied along the main-axis. 
> I am also changing this PR to apply margin-right (layout=row) and margin-bottom (layout=column).

maxime1992:
Ok I skipped that part, my bad.

If there's no `fxLayoutWrap`, it's a good idea.

On the contrary, when the `fxLayoutWrap` is set to `wrap`, devs will have to apply margins manually instead of simply using `fxLayoutGap`. So in this case I'm not so sure.

PR #159, opened 2017-02-05
fix(breakpoints): resolve 1px hole between lg -> xl breakpoints

### LINKED ISSUE #149: 1px hole between lg and xl
All media-query aliases begin there where previous alias ends, e.g. md ends on 1279 px and lg begins at 1280 so if we have fxFlex .xs .md .lg .xl we covered all the possible screen sizes, right? No. because there is one pixel that belongs neither to lg nor to xl (lg ends on 1919px and xl starts at 1921), so if we have let's say
fxFlex=100% fxFlex.sm=70% fxFlex.md=50% fxFlex.lg=40% fxFlex.xl=20% and the screen becomes exactly 1920 pixel wide, our element will get the setting for extra small screen (100%), since 1920 is not large anymore, but not extra large yet. I know that I can use gt-lg instead, but what is the point of xl alias then? If we use xl than we could suddenly run into the difficult to reproduce "bug" for element being too big only if the screen size is exactly 1920 pixel wide. I find the situation confusing, I think the xl should either begin where lg ends or do not exist at all since gt-lg is practically the same. But I would prefer xl and gt-lg be exact aliases. 

Am I missing something (since there is a comment in code with // should be distinct from 'gt-lg' range which points to the intentional difference between lg and xl)?

ThomasBurleson:
* alias `xl` now has mediaQuery == `screen and (min-width: 1920px) and (max-width: 5000px)`

fixes #149.

### DIFF
--- a/src/lib/media-query/breakpoints/break-points.ts
+++ b/src/lib/media-query/breakpoints/break-points.ts
@@ -65,7 +65,7 @@ export const RAW_DEFAULTS: BreakPoint[ ] = [
     alias: 'xl',
     suffix: 'Xl',
     overlapping: false,
-    mediaQuery: 'screen and (min-width: 1921px)'  // should be distinct from 'gt-lg' range
+    mediaQuery: 'screen and (min-width: 1920px) and (max-width: 5000px)'
   }
 ];
 


qaznine:
screen shot 2017-11-01 at 3 32 59 pm

qaznine:
why all gt-xx and lt-xx has different min and max width but gt-lg and lt-xl is 1920px?
is it correct for xl and lt-xl sharing 1920px at the same time?

PR #167, opened 2017-02-07
fix(api): remove use of static ngModule.forRoot()

ThomasBurleson:
BREAKING CHANGE:

Previously releases used FlexLayoutModule.forRoot(). This has been deprecated and removed.

-*before*-

```js
@NgModule({
  declarations : [...],
  imports : [
    CommonModule,
    FlexLayoutModule.forRoot()
  ]
})
export class DemosResponsiveLayoutsModule { }
```

-*after*-

```js
@NgModule({
  declarations : [...],
  imports : [ CommonModule, FlexLayoutModule ]
})
export class DemosResponsiveLayoutsModule { }
```


glebmachine:
Why you remove forRoot? That's cause to multiple instance of MatchMedia services, right?

PR #183, opened 2017-02-14
fix(fxFlex): fxFlex=auto with overlapping breakpoints activated

### LINKED ISSUE #135: Layout with fxFlex="auto" not restoring max-height values properly
**on page load** scans breakpoint overlaps too aggressively [and matches the smallest instead of the largest matching].

```html
<div fxFlex="auto" fxFlex.gt-xs="33%" fxFlex.gt-sm="50%"></div>
```

I'm using fxFlex with 3 breakpoints and IE seems to only pick the first one it passes:

**on page load**, in spite of gt-sm qualifying, it defaults to the gt-xs value. However if I manually re-size the browser it will correct itself. This happens in IE11 only.

ThomasBurleson:
Layout changes from gt-sm will not restore default layout with auto

fix_135

fixes #135.

### DIFF
--- a/src/demo-app/app/github-issues/DemosGithubIssues.ts
+++ b/src/demo-app/app/github-issues/DemosGithubIssues.ts
@@ -3,29 +3,32 @@ import {Component} from '@angular/core';
 @Component({
   selector: 'demos-github-issues',
   template: `
-        <demo-issue-5345></demo-issue-5345>
-        <demo-issue-9897></demo-issue-9897>
-        <demo-issue-181></demo-issue-181>
-    `
+      <demo-issue-5345></demo-issue-5345>
+      <demo-issue-9897></demo-issue-9897>
+      <demo-issue-181></demo-issue-181>
+      <demo-issue-135> </demo-issue-135>
+  `
 })
 export class DemosGithubIssues {
 }
 
-import {NgModule}            from '@angular/core';
-import {CommonModule}        from "@angular/common";
-import {MaterialModule}      from "@angular/material";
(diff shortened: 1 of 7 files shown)


ThomasBurleson:
@d3viant0ne - plz review.

.gitignore
```diff
@@ -18,6 +18,8 @@ build/Release
 
 # Users Environment Variables
 .lock-wscript
+yarn.lock
```
d3viant0ne: As opposed to ignoring the lock file, I would suggest adding a `.gitattributes` file and then treating it as a binary. Removes the merge conflict issue while still providing the ability to easily see any shifting dependencies upstream.

i.e 

```
# .gitattributes
yarn.lock -diff
* text=auto
bin/* eol=lf
```

src/demo-app/app/github-issues/DemosGithubIssues.ts
```diff
@@ -3,29 +3,34 @@ import {Component} from '@angular/core';
...
-import {FlexLayoutModule}    from "../../../lib";     // `gulp build:components` to deploy to node_modules manually
+import {NgModule}         from '@angular/core';
+import {CommonModule}     from "@angular/common";
+import {MaterialModule}   from "@angular/material";
+import {FlexLayoutModule} from "../../../lib";     // `gulp build:components` to deploy to node_modules manually
 
-import {DemoIssue5345}     from "./issue.5345.demo";
-import {DemoIssue9897}     from "./issue.9897.demo";
-import {DemoIssue181}        from './issue.181.demo';
+import {DemoIssue5345}    from "./issue.5345.demo";
```
mmalerba: these first 2 are double imported

PR #195, opened 2017-02-22
fix(build): remove use of Angular private API 

### LINKED ISSUE #181: Wrong layout with fxHide / fxShow sometimes ignoring breakpoint aliases
Hellou,
https://github.com/angular/flex-layout/issues/140 has been marked as solved, but I just found another few errors in `2.0.0-beta.5`, in regards to `fxShow fxHide.md` and the alike. Consider the following example for reproducing:

```html
	<div fxHide fxShow.gt-xs>Type 1, row a</div>
	<div fxHide fxShow.gt-xs>Type 1, row b</div>
	<div fxHide fxShow.gt-xs>Type 1, row c</div>
	<br/>
	<div fxShow fxHide.md>Type 2, row a</div>
	<div fxShow fxHide.md>Type 2, row b</div>
	<div fxShow fxHide.md>Type 2, row c</div>
```

Set your browser to `md` width and refresh the page. You'd expect to see

	Type 1, row a
	Type 1, row b
	Type 1, row c

, but the output is

	Type 1, row a
	Type 1, row b
	
	Type 2, row c

which is wrong on multiple levels. It seems the views fix themselves on update. Meaning when you resize the window width to `gt-md` and back to `md`, `Type 2, row c` will then be correctly hidden. And when you resize the window to `xs` and back to `gt-xs`, the missing `Type 1, row c` pops up.

So, I see wrong initialization and weird interference of multiple elements. I might be missing something.

Best regards,

### LINKED ISSUE #193: Remove private Angular API usage
When testing Angular 4 in google3 we discovered that flex-layout uses private APIs from `@angular/core`. These apis are no longer available in A4. Please remove the usage:

https://github.com/angular/flex-layout/blob/f9f16f5fce2fea05b4c9fdbb1b4d78b977ec1e56/src/lib/flexbox/api/base.ts#L10-L11
https://github.com/angular/flex-layout/blob/f9f16f5fce2fea05b4c9fdbb1b4d78b977ec1e56/src/lib/flexbox/api/base.ts#L95

in google3 we made the following local modification:

before:
```
import {ElementRef, Renderer, OnDestroy} from '@angular/core';

import {__platform_browser_private__} from '@angular/platform-browser';
const getDOM = __platform_browser_private__.getDOM;

import {applyCssPrefixes} from '../../utils/auto-prefixer';

...

    let value = (element.style as any)['display'] || getDOM().getComputedStyle(element)['display'];
```

after:

```
import {ElementRef, Renderer, OnDestroy} from '@angular/core';

import {applyCssPrefixes} from '../../utils/auto-prefixer';

...

    let value = (element.style as any)['display'] || getComputedStyle(element)['display'];
```

please note that the global `window` object has `getComputedStyle` which `getDOM` delegated the call, so this modification is equivalent to the original both at design time and at run time.

ThomasBurleson:
**flex-layout** uses private APIs from @angular/core. These apis are no longer available in A4.

Fixes #193.

### DIFF
--- a/src/lib/flexbox/api/base.ts
+++ b/src/lib/flexbox/api/base.ts
@@ -7,9 +7,6 @@
  */
 import {ElementRef, Renderer, OnDestroy} from '@angular/core';
 
-import {__platform_browser_private__} from '@angular/platform-browser';
-const getDOM = __platform_browser_private__.getDOM;
-
 import {applyCssPrefixes} from '../../utils/auto-prefixer';
 
 import {ResponsiveActivation, KeyOptions} from '../responsive/responsive-activation';
@@ -92,7 +89,7 @@ export abstract class BaseFxDirective implements OnDestroy {
    */
   protected _getDisplayStyle(source?: HTMLElement): string {
     let element: HTMLElement = source || this._elementRef.nativeElement;
-    let value = (element.style as any)['display'] || getDOM().getComputedStyle(element)['display'];
+    let value = (element.style as any)['display'] || getComputedStyle(element)['display'];
     return value.trim();
   }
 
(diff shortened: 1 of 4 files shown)


ThomasBurleson:
@IgorMinar - FYI. And any thoughts on **NOTE** above ?

PR #207, opened 2017-03-08
feat(layout): add wrap options support to fxLayout

ThomasBurleson:


thekalinga:
What happens to `fxLayoutWrap`?

ThomasBurleson:
Deprecated but not removed.

PR #209, opened 2017-03-09
chore(demo): remove unnecessary attributes

amitport:
it is a very confusing mention of a directive that is not actually used in this demo.
Additionally, looks like setting this on router root is a practice likely to change (see https://github.com/angular/material2/issues/1719)

### DIFF
--- a/src/demo-app/app/demo-app/demo-app.ts
+++ b/src/demo-app/app/demo-app/demo-app.ts
@@ -22,7 +22,7 @@ import {Component, ViewEncapsulation} from '@angular/core';
         </div>
     </md-toolbar>
     
-    <div #root="$implicit" dir="ltr" class="demo-content">
+    <div class="demo-content">
       <router-outlet></router-outlet>
     </div>
   `,


ThomasBurleson:
@amitport - this has nothing to do with `@angular/flex-layout`.  **`dir`** is a standard HTML attribute. https://css-tricks.com/almanac/properties/d/direction/

Closing as invalid.

amitport:
@ThomasBurleson 
Of course, I know, but there is no reason to mention it, since the document will be ltr nonetheless.

More importantly, the `#root="$implicit"` doesn't have any effect (AFAIK) and is very confusing. I assume it's only there because it was copied from [material repo](https://github.com/angular/material2/blob/master/src/demo-app/demo-app/demo-app.html#L37)

ThomasBurleson:
@amitport - We misunderstood your intent. Your PR is valid and reasonable... reopening.

ThomasBurleson:
@amitport - please update both the PR and the commit title to conform to our issue/PR conventions:

`chore(demo): remove unnecessary attributes`

amitport:
np, done.

PR #220, opened 2017-03-15
feat(fxFlex): compute immediate parent flex-direction

ThomasBurleson:
Support use of **fxFlex** directive on HTML element without a **fxLayout**  parent
directive. **fxFlex** requires a `flex-direction` to be specific on the immediate parent:

  * use defined flex-direction or getComputed.
  * auto-inject **fxLayout** styles to parent if parent `flex-direction` not already provided.


ThomasBurleson:
@mmalerba - ready for presubmit. Thx.

PR #228, opened 2017-03-19
fix(ngClass,ngStyle): support proper API usages and ChangeDetectionStrategy.OnPush strategies

### LINKED ISSUE #206: Class directive not applied correctly for ngFor and ChangeDetectionStrategy.OnPush
I am trying to use ChangeDetectionStrategy.OnPush along with an ngFor, but have found that the class/ngClass directive does not work when using OnPush.  What I am finding is that the appropriate class is being applied to **only** the last item in the list.

This occurs only on the initial render.  If the screen is resized to cause a change then everything works from that point forward.  I have tried ChangeDetectorRef.markForCheck() and seen no change.  One thing that causes some head scratching is that the class **is** applied to the last item so it's obvious the class directive is firing correctly somewhere.

I have put up a plunker (appropriated from a demo for side nav).  If you look at the file `src/app/responsive-sidenav.ts` . There is a line in the meta that can be commented out to see the "correct" behavior.

http://plnkr.co/edit/zalFNqqiznHUyg33FJII?p=preview

### LINKED ISSUE #215: API Documentation for ngStyle and ngClass uses different syntax
### Bug
In API documentation is written to use `ngClass` and `ngStyle` but it doesn't work. To use responsive features you have to use `class` and `style` 

### Example
To change font-size you have to use:
`[style.sm]="{'font-size.px': 40}"`
instead of:
`[ngStyle.sm]="{'font-size.px': 40}"`

https://plnkr.co/edit/b7XxsdgBAISf0KzRDwte?p=preview

### Version 
@angular/flex-layout: ^2.0.0-rc.1

ThomasBurleson:
* fix(ngClass):  properly differentiate between `ngClass` and `class` api usages
  *  `class.<xxx>` are destructive overrides (as seen in `NgClass::klass`)
  *  `ngClass.<xxx>` use iterables to merge/enable/disable 0...n classes
* fix(ngClass,ngStyle): support ChangeDetectionStrategy.OnPush
  * Add support for both **OnPush** change detection strategies and for @Input changes.

>  @see https://plnkr.co/edit/jCICQ1GXFnzFqFTFlXwh?p=preview

fixes #206, fixes #215.

### DIFF
--- a/src/lib/flexbox/api/base-adapter.ts
+++ b/src/lib/flexbox/api/base-adapter.ts
@@ -1,12 +1,36 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {ElementRef, Renderer} from '@angular/core';
+
 import {BaseFxDirective} from './base';
 import {ResponsiveActivation} from './../responsive/responsive-activation';
 import {MediaQuerySubscriber} from '../../media-query/media-change';
+import {MediaMonitor} from '../../media-query/media-monitor';
+
 
 /**
  * Adapter to the BaseFxDirective abstract class so it can be used via composition.
  * @see BaseFxDirective
  */
 export class BaseFxDirectiveAdapter extends BaseFxDirective {
+
+  /**
+   * Accessor to determine which @Input property is "active"
+   * e.g. which property value will be used.
+   */
+  get activeKey() {
+    let mqa = this._mqActivation;
+    let key = mqa ? mqa.activatedInputKey : this._baseKey;
+    // Note: ClassDirective::SimpleChanges uses 'klazz' instead of 'class' as a key
+    return (key === 'class') ? 'klazz' : key;
+  }
+
+  /** Hash map of all @Input keys/values defined/used */
   get inputMap() {
     return this._inputMap;
   }
@@ -19,10 +43,21 @@ export class BaseFxDirectiveAdapter extends BaseFxDirective {
   }
 
   /**
+   * BaseFxDirectiveAdapter constructor
+   */
+  constructor(protected _baseKey: string,   // non-responsive @Input property name
+              protected _mediaMonitor: MediaMonitor,
+              protected _elementRef: ElementRef,
+              protected _renderer: Renderer ) {
+    super(_mediaMonitor, _elementRef, _renderer);
+  }
+
+
+  /**
    * @see BaseFxDirective._queryInput
    */
   queryInput(key) {
-    return this._queryInput(key);
+    return key ? this._queryInput(key) : undefined;
   }
 
   /**
--- a/src/lib/flexbox/api/class.ts
+++ b/src/lib/flexbox/api/class.ts
@@ -9,18 +9,15 @@ import {
   Directive,
   ElementRef,
   Input,
+  DoCheck,
   OnDestroy,
-  OnInit,
   Renderer,
-  OnChanges,
-  SimpleChanges,
   IterableDiffers,
-  KeyValueDiffers
+  KeyValueDiffers, SimpleChanges, OnChanges
 } from '@angular/core';
 import {NgClass} from '@angular/common';
 
 import {BaseFxDirectiveAdapter} from './base-adapter';
-import {BreakPointRegistry} from './../../media-query/breakpoints/break-point-registry';
 import {MediaChange} from '../../media-query/media-change';
 import {MediaMonitor} from '../../media-query/media-monitor';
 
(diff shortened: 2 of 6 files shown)


src/lib/flexbox/api/class.ts
```diff
@@ -72,7 +69,7 @@ export class ClassDirective extends NgClass implements OnInit, OnChanges, OnDest
   @Input('ngClass.gt-lg') set ngClassGtLg(val: NgClassType) { this._base.cacheInput('classGtLg', val, true); };
 
   /** Deprecated selectors */
-  @Input('class')         set classBase(val: NgClassType) { this._base.cacheInput('class', val, true); }
+  @Input('class')         set klazz(val: NgClassType) { this._base.cacheInput('class', val, true); }
```
ThomasBurleson: This is a similar convention used by NgClass:
```js
  @Input('class')
  set klass(v: string) { ... }
```

ThomasBurleson:
@jelbourn - comments addressed. Ready for presubmit.

ThomasBurleson:
@jelbourn - I will remove support for class.<xxx> and style.<xxx> responsive APIs and require developers to use ngClass.<xxx> and ngStyle.<xxx> APIs.

ThomasBurleson:
@jelbourn - ready for review and presubmit. All tests (including the Angular core `ng-class-spec.ts` tests) pass.

ThomasBurleson:
@mmalerba - can you review these fixes ? Thanks.

src/lib/flexbox/api/class.ts
```diff
@@ -32,112 +29,172 @@ export type NgClassType = string | string[] | Set<string> | {[klass: string]: an
...
-  @Input('ngClass.gt-sm') set ngClassGtSm(val: NgClassType) { this._base.cacheInput('classGtSm', val, true);} ;
-  @Input('ngClass.gt-md') set ngClassGtMd(val: NgClassType) { this._base.cacheInput('classGtMd', val, true);};
-  @Input('ngClass.gt-lg') set ngClassGtLg(val: NgClassType) { this._base.cacheInput('classGtLg', val, true); };
+  @Input('ngClass.xs')    set ngClassXs(val: NgClassType) { this._ngClassAdapter.cacheInput('ngClassXs', val, true); }
+  @Input('ngClass.sm')    set ngClassSm(val: NgClassType) { this._ngClassAdapter.cacheInput('ngClassSm', val, true); };
+  @Input('ngClass.md')    set ngClassMd(val: NgClassType) { this._ngClassAdapter.cacheInput('ngClassMd', val, true); };
+  @Input('ngClass.lg')    set ngClassLg(val: NgClassType) { this._ngClassAdapter.cacheInput('ngClassLg', val, true);};
+  @Input('ngClass.xl')    set ngClassXl(val: NgClassType) { this._ngClassAdapter.cacheInput('ngClassXl', val, true); };
+
+  @Input('ngClass.lt-xs') set ngClassLtXs(val: NgClassType) { this._ngClassAdapter.cacheInput('ngClassLtXs', val, true); };
```
mmalerba: Is `lt-xs` a real thing? I think this should be `lt-xl`

src/lib/flexbox/api/base-adapter.ts
```diff
@@ -1,12 +1,37 @@
...
  * @see BaseFxDirective
  */
 export class BaseFxDirectiveAdapter extends BaseFxDirective {
+
+  /**
+   * Accessor to determine which @Input property is "active"
+   * e.g. which property value will be used.
+   */
+  get activeKey() {
+    let mqa = this._mqActivation;
```
mmalerba: this seems pointless? just use _mqActivation below

src/lib/flexbox/api/base-adapter.ts
```diff
@@ -1,12 +1,37 @@
...
+    let key = mqa ? mqa.activatedInputKey : this._baseKey;
+    // ClassDirective::SimpleChanges uses 'klazz'
+    // instead of 'class' as a key
+    return (key === 'class') ? 'klazz' : key;
```
mmalerba: Just want to double check, this is really what you want right? I see `klazz`, `clazz` and `klass` at various points in the code. Also future work: we should make this consistent everywhere, I vote for `cssClass`

src/lib/flexbox/api/class.ts
```diff
@@ -32,112 +29,172 @@ export type NgClassType = string | string[] | Set<string> | {[klass: string]: an
...
+    this.klass = val;
+  }
+
+  @Input('class.xs')      set classXs(val: NgClassType) { this._classAdapter.cacheInput('classXs', val, true); }
+  @Input('class.sm')      set classSm(val: NgClassType) {  this._classAdapter.cacheInput('classSm', val, true); };
+  @Input('class.md')      set classMd(val: NgClassType) { this._classAdapter.cacheInput('classMd', val, true);};
+  @Input('class.lg')      set classLg(val: NgClassType) { this._classAdapter.cacheInput('classLg', val, true); };
+  @Input('class.xl')      set classXl(val: NgClassType) { this._classAdapter.cacheInput('classXl', val, true); };
+
+  @Input('class.lt-xs')   set classLtXs(val: NgClassType) { this._classAdapter.cacheInput('classLtXs', val, true); };
```
mmalerba: should this be `lt-xl`?

src/lib/flexbox/api/base-adapter.ts
ThomasBurleson: this is a short-cut reference to keep the code clean.

src/lib/flexbox/api/class.ts
ThomasBurleson: `lt-xl`, `lt-lg`, `lt-md`,`lt-sm` are all valid responsive aliases.

ThomasBurleson: I see... the Inputs are not entirely correct. I will fix.

src/lib/flexbox/api/base-adapter.ts
ThomasBurleson: angular `ng-class.ts` uses `klass`. Since they already set a precedent, why not be consistent?

mmalerba: Ok but is this the correct one? because I do see all three of these in the code. More important than what we decide to call it is that it be consistent within our code instead of having 3 different variants. But I don't think we necessarily have to follow the same naming conventions for internal variables as them, this is a totally separate repo.

mmalerba: Oh ok I see this is referencing the one from angular core, no issue for this PR then, just the naming thing to consider in the future

ThomasBurleson:
FYI - A design doc is being prepared to discuss the Responsive API for ngClass, class, ngStyle, and style. Future changes - based on the document - may be presented via 1 or more additional PRs.

PR #239, opened 2017-03-28
fix(fxLayoutAlign): support flex-start and flex-end options

### LINKED ISSUE #232: FxLayoutAlign="flex-end" renders as flex-start
I'm wrapping my html in a div that looks something like `<div class="flex" fxLayout="column" fxLayoutAlign="flex-end center" fxLayoutGap="10px">` but when I go to inspect element, `justify-content` is set to `flex-start` and not `flex-end`

ThomasBurleson:
Current API translates `start` and `end` to `flex-start` and `flex-end` respectively.
The API should also support the raw `flex-<xxx>` values.

Fixes #232.

### DIFF
--- a/src/lib/flexbox/api/layout-align.ts
+++ b/src/lib/flexbox/api/layout-align.ts
@@ -141,10 +141,6 @@ export class LayoutAlignDirective extends BaseFxDirective implements OnInit, OnC
   protected _buildCSS(align) {
     let css = {}, [main_axis, cross_axis] = align.split(' '); // tslint:disable-line:variable-name
 
-    css['justify-content'] = 'flex-start';  // default main axis
-    css['align-items'] = 'stretch';         // default cross axis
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@mmalerba - please review ?

src/lib/flexbox/api/layout-align.ts
```diff
@@ -157,12 +155,22 @@ export class LayoutAlignDirective extends BaseFxDirective implements OnInit, OnC
...
+      case 'flex-end':
         css['justify-content'] = 'flex-end';
         break;
+      case 'start':
+      case 'flex-start':
+      default :
+        break;
     }
+
+    css['align-items'] = 'stretch';         // default cross axis
```
mmalerba: nit: why not just put these in the default block? (same above)

PR #271, opened 2017-04-29
chore(mediaQuery): refactor BreakPointsRegistry to self-register with…

ThomasBurleson:
Refactor based on architecture discussions.

### DIFF
--- a/src/lib/flexbox/api/layout.spec.ts
+++ b/src/lib/flexbox/api/layout.spec.ts
@@ -44,7 +44,7 @@ describe('layout directive', () => {
       imports: [CommonModule, FlexLayoutModule],
       declarations: [TestLayoutComponent],
       providers: [
-        BreakPointRegistry, DEFAULT_BREAKPOINTS_PROVIDER,
+        DEFAULT_BREAKPOINTS_PROVIDER, BreakPointRegistry,
         {provide: MatchMedia, useClass: MockMatchMedia}
       ]
     });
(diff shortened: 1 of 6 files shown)


ThomasBurleson:
@mmalerba:

Based on our architecture discussions, I created this PR. 
Unfortunately it has one major issue:  Testing manifests **cyclic dependency**:

```console
Chrome 58.0.3029 (Mac OS X 10.11.6) layout directive with responsive features should fallback to closest overlapping value when the active mediaQuery change is not configured FAILED
        Error: Provider parse errors:
        Cannot instantiate cyclic dependency! BreakPointRegistry: in NgModule DynamicTestModule in ./DynamicTestModule
            at NgModuleProviderAnalyzer.parse (dist/vendor/@angular/compiler/bundles/compiler.umd.js:10411:19) [ProxyZone]
            at NgModuleCompiler.compile (dist/vendor/@angular/compiler/bundles/compiler.umd.js:16907:24) [ProxyZone]
```
> Try `gulp test:single` in a terminal shell.


The reason is as follows:

* Refactor injects **MatchMedia** into **BreakPointRegistry**
* Testing with `MockMatchMedia extends MatchMedia` injects **BreakPointRegistry**

As such, we get a cyclical injection dependency. And even if MockMatchMedia did not extend MatchMedia, it would still need to inject the BreakPointRegistry in order to support some logic for simulating breakpoint/query activations during testing.

Do to the extensive use of MockMatchMedia and mocks in-general, I do not recommend that we proceed with this refactor as it has 

*  (a) larger downstream impacts, and 
*  (b) introduces complexities with mock services.

I welcome your thoughts on this.

mmalerba:
Whoops sorry forgot to respond to this. I agree that we should avoid doing this for now. I think when we move parts of this to the CDK we should carefully consider what the interface for those parts that will live in the CDK is, and make sure that they will work well for flex-layout and other use cases. Until then it makes sense to hold off on big breaking changes

ThomasBurleson:
@mmalerba - sounds good. Thanks for the follow-up.

PR #276, opened 2017-05-04
Improve flex css prefixes

karlhaas:
With the following changes the layout of our application works in iOS8

### DIFF
--- a/src/lib/flexbox/api/base.ts
+++ b/src/lib/flexbox/api/base.ts
@@ -105,6 +105,19 @@ export abstract class BaseFxDirective implements OnDestroy {
   }
 
   /**
+   * Applies the styles to the element. The styles object map may contain an array of values. Each
+   * value will be added as element style.
+   */
+  protected _applyMultiValueStyleToElement(styles: {}, element: any) {
+    Object.keys(styles).forEach(key => {
+      const values = Array.isArray(styles[key]) ? styles[key] : [styles[key]];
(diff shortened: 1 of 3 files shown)


ThomasBurleson:
@karlhaas - you will need to sign the Google CLA before any merges can occur.

karlhaas:
When I follow the link it tells me "It looks like you've already signed this CLA. If you'd like to edit your contact information, you may do so below."

ThomasBurleson:
@karlhaas - nice work sir!

ThomasBurleson:
@karlhaas - plz squash your commits.

karlhaas:
@ThomasBurleson commits squashed

ThomasBurleson:
@karlhaas  - We are concerned about some performance impacts with this PR.

*  This PR was for iOS8 ? Did this address and any other issues
*  I do not see an Github issue associated with this btw.

karlhaas:
@ThomasBurleson for our application (feedbackr.io) it is very important to support as many devices as possible. According to http://caniuse.com/#feat=flexbox still many browsers need css prefixes for flex.
iOS8 is just one example which I used to test the changes.

Another option (which we don't really want) would not to use flex layout and material for this part of the application.

Could you provide some information regarding the performance impacts? Maybe we can provide a better solution.

ThomasBurleson:
@karlhaas  - see https://github.com/angular/flex-layout/issues/282.

PR #283, opened 2017-05-10
fix(auto-prefixer): resolve perf impacts as reported by LightHouse

### LINKED ISSUE #282: fix(auto-prefixer): resolve perf impacts as reported by LightHouse
*  Order of flex properties should be improved to use `display: flex` last.
*  missing standard property `display: -webkit-flex;`
*  mistakenly using `display: -moz-box;`

See

*  https://github.com/GoogleChrome/lighthouse/issues/1710#issuecomment-294470505
*  https://github.com/GoogleChrome/lighthouse/issues/2022#issuecomment-295022719

ThomasBurleson:
Internal AutoPrefixer updated based on output from prefix rules for n-2 evergreen browsers.

* Updated based on output from https://autoprefixer.github.io/
* Removed Mozilla Firefox prefixes as not required for the evergreen browser
* Remove support for IE10 prefixes (`-ms-`); as only IE 11 and Edge are supported
* Remove support for `-webkit-box` required for OLD iOS , Safari 3, BB7
* Fix value transforms for flex-related keys

Fixes #282

### DIFF
--- a/src/lib/utils/auto-prefixer.ts
+++ b/src/lib/utils/auto-prefixer.ts
@@ -6,27 +6,31 @@
  * found in the LICENSE file at https://angular.io/license
  */
 
-/** Applies CSS prefixes to appropriate style keys.*/
+/**
+ * Applies CSS prefixes to appropriate style keys.
+ *
+ * Note: `-ms-`, `-moz` and `-webkit-box` are no longer supported. e.g.
+ *    {
+ *      display: -webkit-flex;     NEW - Safari 6.1+. iOS 7.1+, BB10
+ *      display: flex;             NEW, Spec - Firefox, Chrome, Opera
+ *      // display: -webkit-box;   OLD - iOS 6-, Safari 3.1-6, BB7
+ *      // display: -ms-flexbox;   TWEENER - IE 10
+ *      // display: -moz-flexbox;  OLD - Firefox
+ *    }
+ */
 export function applyCssPrefixes(target) {
   for (let key in target) {
-
-    let value = target[key];
+    let value = target[key] || "";
 
     switch (key) {
       case 'display':
         if (value === 'flex') {
           target['display'] = [
-            '-webkit-box',
-            '-moz-box',
-            '-ms-flexbox',
             '-webkit-flex',
             'flex'
           ];
         } else if (value === 'inline-flex') {
           target['display'] = [
-            '-webkit-inline-box',
-            '-moz-inline-box',
-            '-ms-inline-flexbox',
             '-webkit-inline-flex',
             'inline-flex'
           ];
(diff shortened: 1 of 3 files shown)


ThomasBurleson:
@jelbourn - fyi

src/lib/utils/auto-prefixer.ts
```diff
@@ -10,24 +10,21 @@
...
+    let boxOrient = toBoxOrient(value);
 
     switch (key) {
       case 'display':
         if (value === 'flex') {
           target['display'] = [
             '-webkit-box',
-            '-moz-box',
-            '-ms-flexbox',
-            '-webkit-flex',
```
jelbourn: I think you removed the wrong one here; you want to keep `-webkit-flex` and remove `-webkit-box`
(same for the other related properties throughout)

ThomasBurleson: @jelbourn - I am very confused here. [AutoPrefixer](https://autoprefixer.github.io/) uses `-webkit-box` and does **NOT** use `-webkit-flex`. Your comment ^ and some research proves that source wrong. 

Here is what I found:

```css
{
 display: -webkit-box;  /* OLD - iOS 6-, Safari 3.1-6, BB7 */
  display: -ms-flexbox;  /* TWEENER - IE 10 */
  display: -webkit-flex; /* NEW - Safari 6.1+. iOS 7.1+, BB10 */
  display: flex;         /* NEW, Spec - Firefox, Chrome, Opera */
}
```

I will restore the `-webkit-flex` settings.  Thx for the oversight.

src/lib/utils/auto-prefixer.ts
```diff
@@ -10,24 +10,21 @@
 export function applyCssPrefixes(target) {
   for (let key in target) {
 
-    let value = target[key];
+    let value = target[key] || "";
+    let boxValue = toBoxValue(value);
+    let boxDirection = toBoxDirection(value);
```
karlhaas: I think toBoxDirection and toBoxOrient will fail with non string parameters. I did a quick test with your changes and it fails with numeric values.

ThomasBurleson: Thx @karlhaas.

src/lib/utils/auto-prefixer.ts
karlhaas: I can confirm that the layout does not work anymore in iOS8. I will check again after the -webkit prefixes have been restored.

ThomasBurleson: @karlhaas  - can you confirm now with the lastest updates to this PR?

ThomasBurleson:
@karlhaas - Can you check these recent changes on iOS8 ? Thx.

src/lib/utils/auto-prefixer.ts
```diff
@@ -35,114 +39,32 @@ export function applyCssPrefixes(target) {
...
-      case 'align-self':
-        target['-ms-flex-item-align'] = toBoxValue(value);
-        target['-webkit-align-self'] = value;
-        break;
-
-      case 'align-content':
-        target['-ms-align-content'] = toAlignContentValue(value);
-        target['-ms-flex-line-pack'] = toAlignContentValue(value);
-        target['-webkit-align-content'] = value;
+        target['order'] = isNaN(value) ? '0' : value;
```
karlhaas: @ThomasBurleson Thx! the layout seems to work. According to https://www.w3schools.com/cssref/css3_pr_order.asp order needs a prefix as well. I would suggest something like 

```
      case 'order':
        target['order'] = target['-webkit-order'] = isNaN(value) ? '0' : value;
```

In our app there are still some problems with material 2 (eg. dialogs). I would try to continue the work on https://github.com/angular/material2/issues/4385 and reduce the prefixes like you did as soon as somebody from the angular 2 team signals that they will accept the changes ...

ThomasBurleson: Done.

ThomasBurleson:
@jelbourn - please presubmit.

ThomasBurleson:
@tinayuangao - rebase is done.

PR #289, opened 2017-05-17
fix(closure-compiler): use Number to cast

alexeagle:
parseInt requires two arguments within Google (the second is the radix)
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt:
"Always specify this parameter"

Casting with Number('1') is better.


ThomasBurleson:
Thx @alexeagle

PR #325, opened 2017-06-23
refactor(api): make sure all code uses Renderer2

matsko:


ThomasBurleson:
@matsko  - Beta.8 used this Renderer2 approach... but caused weird injection issues when injecting BOTH a Renderer and Renderer2 instance. Another approach is needed.

ThomasBurleson:
Replaced with PR #372. Close as will-not-merge.

Issue #364, opened 2017-08-06
Call to this._getFlowDirection(this.parentElement, true); overwrites flex-direction set in stylesheet


ppham27:
Commit #353 will set the flex-direction of the parent container to row when using fxFlex on a child element in this line. In our app, we set the flex-direction to column in a stylesheet, but the method only reads from the style attribute. Thus, the column flex-direction is overwritten.

This is where the child element sets the flex-direction of the parent element.

https://github.com/angular/flex-layout/blob/master/src/lib/flexbox/api/flex.ts#L162

My workaround has been to do

```js
@Component({                                                                                       
  moduleId: module.id,                                                                             
  selector: 'my-app',                                                                       
  host: {                                                                                          
    style: 'flex-direction: column',                                                               
  },                                                                                               
  styleUrls: ['./my_app.css'],                                                              
  templateUrl: './my_app.ng.html',                                                          
}) 
```

rosslavery:
I believe this is a duplicate of #272.

ThomasBurleson:
Fixed with SHA dde889d

Issue #296, opened 2017-05-25
feat(breakpoints): only apply flex properties if a breakpoint is activated


jeffbcross:
(opening issue per @ThomasBurleson suggestion in #201)

My comment:

> I'd like to be able disable flex layout completely for some elements when the width is extra small. But right now, even if I use gt-xs breakpoints, flex CSS properties are being added to the elements.
>
> 
> ```html
> <div fxLayoutAlign.gt-xs="center" fxLayout.gt-xs="row" fxLayoutGap.gt-xs="25px">
>   <div fxFlex.gt-xs="500px" class="victor"></div>
> </div>
> ```
>
> The layout element has `display: flex; flex-direction: row;` and the "victor" element has `flex: 1 1 0 0.00000001px;` when the width is xs. I expected that using only `gt-xs`, and no defaults, would mean that the flex properties would only be applied when the screen was greater than xs.

For further context, the main motivation is for a responsive layout where I want rows to be columns (stacked) on mobile. Safari has requirement of parent column containers having a set height (not auto), or else elements start clobbering each other. I want the child elements to be `display: block; height: auto;`.

Thanks :)

jeffbcross:
BTW happy to submit a PR if maintainers agree with the request and we can agree on a design, but I'm guessing this would require significant internal changes.

ThomasBurleson:
@jeffbcross - When using the flex-layout API, if you use responsive directives WITHOUT default flex-layout directives, then those defaults are **injected** as needed.

So - in your scenario above - the following is what is actually happening:

```html
<div fxLayout="row"
     fxLayoutAlign="start start"
     fxLayoutGap="0px"
     
     fxLayout.gt-xs="row" 
     fxLayoutAlign.gt-xs="center" 
     fxLayoutGap.gt-xs="25px">

  <div fxFlex
       fxFlex.gt-xs="500px" 
       class="victor"></div>

</div>
```

Now if you have CSS define for `flex-direction:column` then the default fxLayout will use that...
We presume defaults because those values are restored when a responsive breakpoint **de-activates".

Closing as non-issue.

ThomasBurleson:
@jeffbcross - The current architecture assumption is that any responsive API usage implies that the host element will have default (non-responsive) flexbox settings auto-applied . 

Under what scenarios would you want the container `box` to NOT have default flexbox settings when you specified such for mobile? 

```
<div class="box" fxLayout.gt-xs="row"></div>
```

[...]

PR #365, opened 2017-08-06
fix(flexbox): scan flex-direction in css stylesheet

### LINKED ISSUE #272: Parent container flexbox injection logic is wrong
When a Flex Layout directive is used on a child element, it expects the parent to work as a FlexBox container. If the parent does not have "display:flex" stylings, then these are injected, inline onto the parent.

If the parent has a classname that defines the FlexBox stylings, these are ignored as only the inline styles are checked. This is wrong.

ThomasBurleson:
`fxFlex` works best if the parent container has flexbox styles assigned (display, flex-direction, etc.). If not assigned (inline or via stylesheet), then these styles are auto-injected inline on the DOM parent element.

When scanning parent DOM element for flexbox flow direction, first scan the in element's inline style and then scan the computed stylesheet for the `flex-direction` value.

> It is assumed that if the parent element has a `flex-direction` style property set, then the other required properties (display, border-box, etc.)  have also been assigned.

Fixes #272, #364.

### DIFF
--- a/src/lib/flexbox/api/base.ts
+++ b/src/lib/flexbox/api/base.ts
@@ -18,7 +18,6 @@ import {ResponsiveActivation, KeyOptions} from '../responsive/responsive-activat
 import {MediaMonitor} from '../../media-query/media-monitor';
 import {MediaQuerySubscriber} from '../../media-query/media-change';
 
-
 /**
  * Definition of a css style. Either a property name (e.g. "flex-basis") or an object
  * map of property name and value (e.g. {display: 'none', flex-order: 5}).
@@ -120,9 +119,16 @@ export abstract class BaseFxDirective implements OnDestroy, OnChanges {
     let element: HTMLElement = source || this._elementRef.nativeElement;
     let value = this._lookupStyle(element, 'display');
 
-    return value ? value.trim() : ((element.nodeType === 1) ? 'block' : 'inline-block');
+    return value ? value.trim() :
+        ((element.nodeType === 1) ? 'block' : 'inline-block');
   }
 
+  /**
+   * Determine the DOM element's Flexbox flow (flex-direction).
+   *
+   * Check inline style first then check computed (stylesheet) style.
+   * And optionally add the flow value to element's inline style.
+   */
   protected _getFlowDirection(target: any, addIfMissing = false): string {
     let value = 'row';
 
@@ -146,7 +152,7 @@ export abstract class BaseFxDirective implements OnDestroy, OnChanges {
     try {
       if (element) {
         let immediateValue = getDom().getStyle(element, styleName);
-        value = immediateValue || getDom().getComputedStyle(element).display;
+        value = immediateValue || getDom().getComputedStyle(element).getPropertyValue(styleName);
       }
     } catch (e) {
       // TODO: platform-server throws an exception for getComputedStyle
--- a/src/lib/flexbox/api/flex.spec.ts
+++ b/src/lib/flexbox/api/flex.spec.ts
@@ -118,6 +118,26 @@ describe('flex directive', () => {
       expect(element).not.toHaveCssStyle({'min-width': '30px'});
     });
 
+    it('should CSS stylesheet and not inject flex-direction on parent', () => {
+      componentWithTemplate(`
+        <style>
+          .test { flex-direction:column; display: flex; }
+        </style>
+        <div class='test'>
+          <div fxFlex='30px' fxFlex.gt-sm='50'  >  </div>
+        </div>
+      `);
+
+      fixture.detectChanges();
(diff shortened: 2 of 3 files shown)


ThomasBurleson:
@andrewseguin  - can you review ? Thx.

src/lib/flexbox/api/base.ts
```diff
@@ -146,7 +152,7 @@ export abstract class BaseFxDirective implements OnDestroy, OnChanges {
     try {
       if (element) {
         let immediateValue = getDom().getStyle(element, styleName);
-        value = immediateValue || getDom().getComputedStyle(element).display;
+        value = immediateValue || getDom().getComputedStyle(element)[styleName];
```
crisbeto: Instead of `getDom().getComputedStyle(element)[styleName]` you can use `getDom().getComputedStyle(element).getPropertyValue(styleName)`.

src/lib/flexbox/api/flex.ts
```diff
@@ -77,10 +77,14 @@ export class FlexDirective extends BaseFxDirective implements OnInit, OnChanges,
...
   @Input('fxFlex.lt-xl') set flexLtXl(val) { this._cacheInput('flexLtXl', val); };
-
   /* tslint:enable */
-  // Explicitly @SkipSelf on LayoutDirective and LayoutWrapDirective because we want the
-  // parent flex container for this flex item.
+
+  /**
+   * Constructor
+   *
+   * Note: Explicitly @SkipSelf on LayoutDirective and LayoutWrapDirective because we are looking
```
crisbeto: This note might be better as a regular comment, otherwise it'll show up in any generated docs and typings.

src/lib/flexbox/api/flex.spec.ts
```diff
@@ -118,6 +118,26 @@ describe('flex directive', () => {
       expect(element).not.toHaveCssStyle({'min-width': '30px'});
     });
 
+    it('should CSS stylesheet and not inject flex-direction on parent', () => {
+      componentWithTemplate(`
+        <style type="text/css">
```
crisbeto: The `text/css` shouldn't be necessary.

src/lib/flexbox/api/base.ts
```diff
@@ -123,6 +123,12 @@ export abstract class BaseFxDirective implements OnDestroy, OnChanges {
     return value ? value.trim() : ((element.nodeType === 1) ? 'block' : 'inline-block');
```
crisbeto: Not something that was added in this PR, but instead of hardcoding the `1`, you can use the built-in browser constant: `element.nodeType === Node.ELEMENT_NODE`.

src/lib/flexbox/api/base.ts
ThomasBurleson: IYO, which is preferred/better ?

src/lib/flexbox/api/flex.ts
ThomasBurleson: Good point. Thx.

src/lib/flexbox/api/flex.spec.ts
ThomasBurleson: k. will remove.

src/lib/flexbox/api/base.ts
ThomasBurleson: sgtm. Thx.

src/lib/flexbox/api/base.ts
crisbeto: I believe `getPropertyValue` guarantees that the style was recalculated.

ThomasBurleson: Good to know. Thx.

ThomasBurleson:
@crisbeto using `Node.ELEMENT_NODE` causes problems with angular universal compiles.

matheusdavidson:
Sorry for saying this but i think you guys are trying to apply styles where you don't need to. This is causing many issues and non expected behaviors.

I'm saying this because i had some problems already with things like using ```fxFlex.gt-xs``` and flex layout adding styles even when breakpoint is under ```xs```(talked about this in another issue #296).

@ThomasBurleson, you said: "fxFlex works best if the parent container has flexbox styles assigned.."

I ask you: Wouldn't it be better to let the user choose if he wants a flexbox parent?

Really, i think that worse then having a ```fxFlex``` without flexbox parent(this obvious) is to have no idea why my parent element got flexbox in the first place, since i was already applying it in my css file and flex layout just override the whole thing.

src/lib/flexbox/api/base.ts
Matrosskin: As I know, in iOS 9 the `getPropertyValue` returns `null` instead of empty string as in browsers on other OS. Not sure how much people still use the iOS 9 so my comment is just for info.

PR #367, opened 2017-08-07
fix(breakpoints): support print media

### LINKED ISSUE #361: Print media support needed
Flex-layout responsive breakpoints not working on print media. 

My pages look awful on `@media print`. Initial investigation reveals that at least `fxFlex.gt-xs="..."` isn't working when emulating CSS print media. What is this project's stance on supporting print media? I haven't seen anything in the docs.

[Here's a quick plunkr](https://plnkr.co/edit/5Cs8QhRb48sYzk2FqVP2?p=preview). Make the preview big to activate `gt-xs`, and then emulate print CSS media in Chrome.

ThomasBurleson:
Previous breakpoints defined mediaQueries like `screen and (max-width: 599px)`. This means that
all responsive layouts (using .xs, .sm, etc.) will not print properly since `@media print` is excluded
by the `screen and ...` breakpoint definitions.

Remove all `screen and ` @media prefixes; so the layouts will appear the same on both display and print devices.

Fixes #361.

### DIFF
--- a/src/lib/media-query/breakpoints/data/break-points.ts
+++ b/src/lib/media-query/breakpoints/data/break-points.ts
@@ -14,63 +14,63 @@ export const RESPONSIVE_ALIASES = [
 export const DEFAULT_BREAKPOINTS: BreakPoint[] = [
   {
     alias: 'xs',
-    mediaQuery: 'screen and (max-width: 599px)'
+    mediaQuery: '(max-width: 599px)'
   },
   {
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@mmalerba - this actually passed on Travis... not sure **why** it is show failed when it all passed. 

> https://travis-ci.org/angular/flex-layout/builds/261842949?utm_source=github_status&utm_medium=notification

PR #369, opened 2017-08-07
fix(fxFlexOffset): use parent flow direction for margin property

### LINKED ISSUE #328: Unexpected behaviour: offset is always horizontal
Hi,
I found an unexpected behaviour when you set fxFlexOffset in a `fxLayout="column"` child, it result in a **margin-left** added; instead of a **margin-top** which is what I was expecting.

Demo here: https://plnkr.co/edit/PDOtQk?p=preview

Is it on purpose or a bug ? Thanks in advance.

ThomasBurleson:
`fxFlexOffset` assigns inline `margin-left` style; assuming default flow directions == 'row'. If the parent flow direction == 'column', then a `margin-top` should be used.

Also added a subscription to an optional parent element LayoutDirective; which will trigger the FlexOffsetDirective to update the inline styles to match the current flow direction.

> Note: if `fxFlexOffset` is used **without**  a parent flexbox styling (set via css or directive), then a `display:flex; flex-direction:row;` will be auto assigned to the fxFlexOffset host element's parent.

Fixes #328.

### DIFF
--- a/src/lib/flexbox/api/flex-offset.spec.ts
+++ b/src/lib/flexbox/api/flex-offset.spec.ts
@@ -1,0 +1,159 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {Component} from '@angular/core';
+import {CommonModule} from '@angular/common';
+import {ComponentFixture, TestBed, inject} from '@angular/core/testing';
+
+import {DEFAULT_BREAKPOINTS_PROVIDER} from '../../media-query/breakpoints/break-points-provider';
+import {BreakPointRegistry} from '../../media-query/breakpoints/break-point-registry';
+import {MockMatchMedia} from '../../media-query/mock/mock-match-media';
+import {MatchMedia} from '../../media-query/match-media';
+import {FlexLayoutModule} from '../../module';
+
+import {customMatchers, expect} from '../../utils/testing/custom-matchers';
+import {_dom as _} from '../../utils/testing/dom-tools';
+
+import {
+  makeExpectDOMFrom,
+  makeCreateTestComponent,
+  queryFor
+} from '../../utils/testing/helpers';
+
+describe('flex directive', () => {
+  let fixture: ComponentFixture<any>;
+  let matchMedia: MockMatchMedia;
+  let expectDOMFrom = makeExpectDOMFrom(() => TestFlexComponent);
+  let componentWithTemplate = (template: string) => {
+    fixture = makeCreateTestComponent(() => TestFlexComponent)(template);
+
+    inject([MatchMedia], (_matchMedia: MockMatchMedia) => {
+      matchMedia = _matchMedia;
+    })();
+  };
+
+  beforeEach(() => {
+    jasmine.addMatchers(customMatchers);
+
+    // Configure testbed to prepare services
+    TestBed.configureTestingModule({
+      imports: [CommonModule, FlexLayoutModule],
+      declarations: [TestFlexComponent],
+      providers: [
+        BreakPointRegistry, DEFAULT_BREAKPOINTS_PROVIDER,
+        {provide: MatchMedia, useClass: MockMatchMedia}
+      ]
+    });
+  });
+
+  describe('with static features', () => {
+
+    it('should add correct styles for default `fxFlexOffset` usage', () => {
+      componentWithTemplate(`<div fxFlexOffset='32px' fxFlex></div>`);
+      fixture.detectChanges();
+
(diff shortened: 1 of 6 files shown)


ThomasBurleson:
@crisbeto - would you mind reviewing these?

src/lib/flexbox/api/flex-offset.spec.ts
```diff
@@ -0,0 +1,163 @@
...
+
+    it('should add correct styles for default `fxFlexOffset` usage', () => {
+      componentWithTemplate(`<div fxFlexOffset='32px' fxFlex></div>`);
+      fixture.detectChanges();
+
+      let dom = fixture.debugElement.children[0].nativeElement;
+      let isBox = _.hasStyle(dom, 'margin-left', '32px');
+      let hasFlex = _.hasStyle(dom, 'flex', '1 1 1e-09px') ||         // IE
+          _.hasStyle(dom, 'flex', '1 1 1e-9px') ||          // Chrome
+          _.hasStyle(dom, 'flex', '1 1 0.000000001px') ||   // Safari
```
crisbeto: Perhaps this should be rounded down in the implementation for consistency?

src/lib/flexbox/api/flex-offset.spec.ts
```diff
@@ -0,0 +1,163 @@
...
+          _.hasStyle(dom, 'flex', '1 1 0px');
+
+      expect(isBox).toBeTruthy();
+      expect(hasFlex).toBeTruthy();
```
crisbeto: Since `hasStyle` returns a boolean, you can `expect(hasFlex).toBe(true)`.

src/lib/flexbox/api/flex-offset.spec.ts
```diff
@@ -0,0 +1,163 @@
...
+      let element = queryFor(fixture, '[fxFlex]')[0].nativeElement;
+      let parent = queryFor(fixture, '.parent')[0].nativeElement;
+
+      // parent flex-direction found with 'column'; set child with height styles
+      expect(element).toHaveCssStyle({'margin-top': '21%'});
+      expect(parent).toHaveCssStyle({'flex-direction': 'column'});
+    });
+
+
+    describe('', () => {
```
crisbeto: This one's missing a title. If the test inside is standalone, they can be moved outside.

src/lib/flexbox/api/flex-offset.ts
```diff
@@ -71,6 +80,16 @@ export class FlexOffsetDirective extends BaseFxDirective implements OnInit, OnCh
   }
 
   /**
+   * Cleanup
+   */
+  ngOnDestroy() {
+    super.ngOnDestroy();
```
crisbeto: AFAIK this isn't necessary as of Angular 2.4.

src/lib/flexbox/api/flex-offset.ts
```diff
@@ -84,7 +103,43 @@ export class FlexOffsetDirective extends BaseFxDirective implements OnInit, OnCh
   // Protected methods
   // *********************************************
 
+  /** The flex-direction of this element's host container. Defaults to 'row'. */
+  protected _layout = 'row';
```
crisbeto: This one should be strongly typed.

src/lib/flexbox/api/flex-offset.spec.ts
ThomasBurleson: required. will not change.

src/lib/flexbox/api/flex-offset.spec.ts
ThomasBurleson: fixed.

src/lib/flexbox/api/flex-offset.spec.ts
ThomasBurleson: done.

src/lib/flexbox/api/flex-offset.ts
ThomasBurleson: will leave as explicit indicator to call overriden, super class method.

src/lib/flexbox/api/flex-offset.ts
ThomasBurleson: Why? I thought explicit value assignment defines type and negates need to declare it.

crisbeto: I think that this one will be inferred to `string` which could be anything. Aren't the only possible values `row` and `column`?

ThomasBurleson: Now I understand. We do have value validators. I think you found a potential bug since we are not using that here NOR checking for row-reverse or column-reverse.

I will fix.

ThomasBurleson: Fixed with `isFlowHorizontal()` call on line 162.

PR #370, opened 2017-08-08
fix(fxLayoutGap): mutation observer should run outside the ngZone

### LINKED ISSUE #329: Strange bug: when using fxLayoutGap on container and *ngFor on child
Hi,

I found a strange bug which, I presume, has something to do with fxLayoutGap directive.

My use case was to use a component method to provide items for a *ngFor loop.
I noticed that if the parent of the loop has the directive fxLayoutGap, browser just crash (tested on Chrome and Firefox).
Demo here: https://plnkr.co/edit/paJ7Sb

Hoping you can help me on this.

Thanks in advance.

ThomasBurleson:
Inside the ngZone, the mutation observer and _updateWithValue() style changes causes an infinite recursion. MutationObserver handlers should be executed outside the ngZone.

Fixes #329.

### DIFF
--- a/src/lib/flexbox/api/layout-gap.ts
+++ b/src/lib/flexbox/api/layout-gap.ts
@@ -16,6 +16,7 @@ import {
   AfterContentInit,
   Optional,
   OnDestroy,
+  NgZone,
 } from '@angular/core';
 import {Subscription} from 'rxjs/Subscription';
 
@@ -65,7 +66,8 @@ export class LayoutGapDirective extends BaseFxDirective implements AfterContentI
   constructor(monitor: MediaMonitor,
               elRef: ElementRef,
               renderer: Renderer,
-              @Optional() @Self() container: LayoutDirective) {
+              @Optional() @Self() container: LayoutDirective,
+              private _zone: NgZone) {
     super(monitor, elRef, renderer);
 
     if (container) {  // Subscribe to layout direction changes
@@ -114,22 +116,23 @@ export class LayoutGapDirective extends BaseFxDirective implements AfterContentI
    * NOTE: this does NOT! differentiate between viewChildren and contentChildren
    */
   protected _watchContentChanges() {
-    let onMutationCallback = (mutations) => {
-      let validatedChanges = (it: MutationRecord) => {
-        return (it.addedNodes && it.addedNodes.length) ||
-            (it.removedNodes && it.removedNodes.length);
-      };
+    this._zone.runOutsideAngular(() => {
 
-      // update gap styles only for child 'added' or 'removed' events
-      if (mutations.filter(validatedChanges).length) {
-        this._updateWithValue();
+      if (typeof MutationObserver !== 'undefined') {
+        this._observer = new MutationObserver((mutations: MutationRecord[]) => {
+          let validatedChanges = (it: MutationRecord): boolean => {
+            return (it.addedNodes && it.addedNodes.length > 0) ||
+                (it.removedNodes && it.removedNodes.length > 0);
+          };
+
+          // update gap styles only for child 'added' or 'removed' events
+          if (mutations.some(validatedChanges)) {
+            this._updateWithValue();
+          }
+        });
+        this._observer.observe(this._elementRef.nativeElement, {childList: true});
       }
-    };
-
-    if (typeof MutationObserver !== 'undefined') {
-      this._observer = new MutationObserver(onMutationCallback);
-      this._observer.observe(this._elementRef.nativeElement, {childList: true});
-    }
+    });
   }
 
   /**


ThomasBurleson:
@crisbeto - would you mind reviewing these?

src/lib/flexbox/api/layout-gap.ts
```diff
@@ -15,7 +15,7 @@ import {
   Self,
   AfterContentInit,
   Optional,
-  OnDestroy,
+  OnDestroy, NgZone,
```
crisbeto: This one should be on the next line.

src/lib/flexbox/api/layout-gap.ts
```diff
@@ -114,22 +115,25 @@ export class LayoutGapDirective extends BaseFxDirective implements AfterContentI
...
    */
   protected _watchContentChanges() {
-    let onMutationCallback = (mutations) => {
-      let validatedChanges = (it: MutationRecord) => {
-        return (it.addedNodes && it.addedNodes.length) ||
-            (it.removedNodes && it.removedNodes.length);
+    this._zone.runOutsideAngular(() => {
+
+      let onMutationCallback = (mutations) => {
+        let validatedChanges = (it: MutationRecord) => {
```
crisbeto: Since this is only used once it can be inlined, but I guess that's down to preference. Same goes for the onMutationCallback.

src/lib/flexbox/api/layout-gap.ts
```diff
@@ -114,22 +115,25 @@ export class LayoutGapDirective extends BaseFxDirective implements AfterContentI
...
+          return (it.addedNodes && it.addedNodes.length) ||
+              (it.removedNodes && it.removedNodes.length);
+        };
+
+        // update gap styles only for child 'added' or 'removed' events
+        if (mutations.filter(validatedChanges).length) {
```
crisbeto: Instead of using `mutation.filter(...).length`, you can use `mutations.some(...)` which will exit earlier.

src/lib/flexbox/api/layout-gap.ts
```diff
@@ -114,22 +115,25 @@ export class LayoutGapDirective extends BaseFxDirective implements AfterContentI
...
+          this._updateWithValue();
+        }
       };
 
-      // update gap styles only for child 'added' or 'removed' events
-      if (mutations.filter(validatedChanges).length) {
-        this._updateWithValue();
+      if (typeof MutationObserver !== 'undefined') {
+        this._observer = new MutationObserver(onMutationCallback.bind(this));
```
crisbeto: The `bind(this)` shouldn't be necessary since it's an arrow function.

src/lib/flexbox/api/layout-gap.ts
crisbeto: I believe this is a `MutationRecord[]`, not a single `MutationRecord`.

src/lib/flexbox/api/layout-gap.ts
ThomasBurleson: done.

src/lib/flexbox/api/layout-gap.ts
ThomasBurleson: fixed.

src/lib/flexbox/api/layout-gap.ts
ThomasBurleson: done.

src/lib/flexbox/api/layout-gap.ts
ThomasBurleson: improved.

src/lib/flexbox/api/layout-gap.ts
ThomasBurleson: removed.

ThomasBurleson:
@crisbeto - nice feedback; as usually. Changes complete.

src/lib/flexbox/api/layout-gap.ts
```diff
@@ -114,22 +116,23 @@ export class LayoutGapDirective extends BaseFxDirective implements AfterContentI
...
-        return (it.addedNodes && it.addedNodes.length) ||
-            (it.removedNodes && it.removedNodes.length);
-      };
-
-      // update gap styles only for child 'added' or 'removed' events
-      if (mutations.filter(validatedChanges).length) {
-        this._updateWithValue();
+    this._zone.runOutsideAngular(() => {
+
+      if (typeof MutationObserver !== 'undefined') {
```
crisbeto: Not a big deal, but you can move the `MutationObserver` support check outside of the `runOutsideAngular` to avoid hitting `runOutsideAngular` completely if it's unsupported.

ThomasBurleson: true. minor nit. agreed ?

crisbeto: Yeah it's pretty minor, no need to change it.

PR #372, opened 2017-08-09
refactor(api): use Renderer2 instead of Renderer

### LINKED ISSUE #326: Flex-layout needs to upgrade to Renderer2
There have been some recent refactorings in angular/angular that change NgClass and other directives to use Renderer2 instead of Renderer. While this change is transparent to the public API of Angular, it is still a change that causes issues to code that extends the base directives that Angular has to offer.

With flex-layout, the NgClass, NgStyle and other directives are extended. This is an anti-pattern and isn't encouraged by Angular itself. Therefore there are two things that need to be changed in flex-layouts so that it's compatible with future versions of Angular.

1) Use Renderer2 instead of Renderer
2) Do not extend the Angular base directives, but instead inject them into the constructor.

Once this code changes have been made then flex-layouts can work with both Angular4 and Angular5.

ThomasBurleson:
* Use Renderer2 instead of Renderer
* Do not extend the Angular base directives, but instead inject them into the constructor.
* improve support for future releases of Angular 5

Fixes #326.

### DIFF
--- a/src/demo-app/app/docs-layout-responsive/responsiveStyle.demo.ts
+++ b/src/demo-app/app/docs-layout-responsive/responsiveStyle.demo.ts
@@ -49,13 +49,14 @@ import {Component} from '@angular/core';
 
       <md-card-content>
         <div class="containerX">
-          <div fxLayout="row" fxFlex class="coloredContainerX box">
-            <div
-              fxFlex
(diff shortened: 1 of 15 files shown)


ThomasBurleson:
@crisbeto - Would you review plz?

PR #374, opened 2017-08-10
fix(api): defer getComputedStyle() calls until ngOnInit phase

### LINKED ISSUE #310: fxShow is setting display:inline; in IE11 and Edge
https://plnkr.co/edit/lF0h5ttXwFiOLrjIgKDr?p=preview

Discovered when using fxShow with md-menu-items. In the above plunkr, the last 4 items in the menu have `fxShow` on them. In IE11 and Edge, `display:inline` is applied to them, which ruins the layout.

ThomasBurleson:
Previously, the Show/Hide directives would query the `display` style during directive instantiation.

For elements instantiated during `*ngFor`, constructor calls to getComputedStyle() will incorrectly return “” due to failures in queries for styles in CSS stylesheets or inline styles. Deferring these types of queries to the ngOnInit() lifecycle event, enables valid response to queries for both inline styles and stylesheets.

Fixes #310.

### DIFF
--- a/src/lib/flexbox/api/base.ts
+++ b/src/lib/flexbox/api/base.ts
@@ -9,20 +9,19 @@ import {
   ElementRef, OnDestroy, SimpleChanges, OnChanges,
   SimpleChange, Renderer2
 } from '@angular/core';
-import {ɵgetDOM as getDom} from '@angular/platform-browser';
 
-import {applyCssPrefixes} from '../../utils/auto-prefixer';
 import {buildLayoutCSS} from '../../utils/layout-validator';
+import {
+  StyleDefinition,
+  lookupStyle,
+  lookupInlineStyle,
+  applyStyleToElement,
+  applyStyleToElements
+} from '../../utils/style-utils';
 
 import {ResponsiveActivation, KeyOptions} from '../responsive/responsive-activation';
 import {MediaMonitor} from '../../media-query/media-monitor';
 import {MediaQuerySubscriber} from '../../media-query/media-change';
-
-/**
- * Definition of a css style. Either a property name (e.g. "flex-basis") or an object
- * map of property name and value (e.g. {display: 'none', flex-order: 5}).
- */
-export type StyleDefinition = string | { [property: string]: string | number };
 
 /** Abstract base class for the Layout API styling directives. */
 export abstract class BaseFxDirective implements OnDestroy, OnChanges {
@@ -66,7 +65,6 @@ export abstract class BaseFxDirective implements OnDestroy, OnChanges {
   constructor(protected _mediaMonitor: MediaMonitor,
               protected _elementRef: ElementRef,
               protected _renderer: Renderer2) {
-    this._display = this._getDisplayStyle();
   }
 
   // *********************************************
@@ -91,6 +89,15 @@ export abstract class BaseFxDirective implements OnDestroy, OnChanges {
   // *********************************************
   // Lifecycle Methods
   // *********************************************
+
+  /**
+   * Use post-component-initialization event to perform extra
+   * querying such as computed Display style
+   */
+  ngOnInit() {
+    this._display = this._getDisplayStyle();
+    this._hasInitialized = true;
+  }
 
   ngOnChanges(change: SimpleChanges) {
     throw new Error(`BaseFxDirective::ngOnChanges should be overridden in subclass: ${change}`);
@@ -124,10 +131,7 @@ export abstract class BaseFxDirective implements OnDestroy, OnChanges {
    */
   protected _getDisplayStyle(source?: HTMLElement): string {
     let element: HTMLElement = source || this._elementRef.nativeElement;
-    let value = this._lookupStyle(element, 'display');
-
-    // Note: 'inline' is the default of all elements, unless UA stylesheet overrides
-    return value ? value.trim() : 'inline';
+    return lookupStyle(element, 'display');
   }
 
   /**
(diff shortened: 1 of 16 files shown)


ThomasBurleson:
@crisbeto - would you mind reviewing this ?

src/lib/flexbox/api/base.ts
```diff
@@ -92,6 +91,15 @@ export abstract class BaseFxDirective implements OnDestroy, OnChanges {
   // Lifecycle Methods
   // *********************************************
 
+  /**
+   * Use post-component-initialization event to perform extra
+   * querying such as computed Display style
+   */
+  ngOnInit() {
+    this._display = this._getDisplayStyle();
+    this._hasInitialized = true;
```
crisbeto: `hasInitialized` is only used in a `ngDoCheck` hook. [According to this chart](https://angular.io/guide/lifecycle-hooks), it shouldn't be possible for `ngDoCheck` to fire before `ngOnInit`, making the flag redundant.

src/lib/flexbox/api/base.ts
```diff
@@ -127,7 +135,7 @@ export abstract class BaseFxDirective implements OnDestroy, OnChanges {
     let value = this._lookupStyle(element, 'display');
 
     // Note: 'inline' is the default of all elements, unless UA stylesheet overrides
-    return value ? value.trim() : 'inline';
+    return value ? value.trim() : ((element.nodeType === 1) ? 'block' : 'inline');
```
crisbeto: Not a huge of nested ternaries. Consider breaking it out into an `if` and a ternary? E.g.

```
if (value) {
  return value;
}

return ((element.nodeType === 1) ? 'block' : 'inline');
```

crisbeto: I'm not sure that checking that the `nodeType === 1` is a valid way to verify that the element is block-level. 1 only means that it's an `Element` node which is true for almost everything except cases like comments or raw text nodes.

ThomasBurleson: Hmmm. That is true. Any suggestions? 

Perhaps, we should KISS with ` return value ? value.trim() : 'block'` ?

src/lib/flexbox/api/base.ts
ThomasBurleson: Not sure what you mean. Many of the directives using `ngOnInit()`, the hasInitialized() is a custom method used in the Flex directives.

src/lib/flexbox/api/base.ts
crisbeto: That won't be true necessarily for all cases. One potential solution can be to have a list of the default block-level elements and verify it against the `element.nodeName`, assuming that everything else is `inline`.

ThomasBurleson: That does not seem reasonable for this library.

src/lib/flexbox/api/base.ts
crisbeto: I mean that the only place I see it being used is https://github.com/angular/flex-layout/pull/374/files#diff-5bc48b81f8309cb16bb22da0a329e466R140 which shouldn't be possible to fire before `ngOnInit`. I might be missing some context here, though.

ThomasBurleson: Currenty `this.hasInitialized` is only used in `ShowHideDirective::ngOnIniti()`

src/lib/flexbox/api/base.ts
ThomasBurleson: `_lookupStyle()` checks in-line styles or call `getComputedStyle()`. So the expected response usually be defined.  

I think `return value ? value.trim() : 'block'` is a reasonable fallback when `_lookupStyle()` fails.

src/lib/flexbox/api/base.ts
ThomasBurleson: `ngOnChanges()` is called 1x before ngOnInit() and then later when the @Input values are externally changed. The intention of that logic is to SKIP any responses to changes until after `ngOnInit()`.

crisbeto: Alright, makes sense. I suppose it also depends on the component tree.

src/lib/flexbox/api/base.ts
crisbeto: It will catch most cases, but not the ones that are set to `block` by the UA, which are:

p
h1, h2, h3, h4, h5, h6
ol, ul
pre
address
blockquote
dl
div
fieldset
form
hr
noscript
table

ThomasBurleson:
@crisbeto - working on 1-2 unit tests.

ThomasBurleson:
@crisbeto - unit-tests provided. Can you review last bits?

src/lib/utils/style-utils.ts
```diff
@@ -0,0 +1,91 @@
...
+  return getDom().getStyle(element, styleName);
+}
+
+/**
+ * Determine the inline or inherited CSS style
+ */
+export function lookupStyle(element: HTMLElement, styleName: string, inlineOnly = false): any {
+  let value = '';
+  if (element) {
+    try {
```
crisbeto: AFAIK try/catch has a slight performance overhead. This one could be worked around by checking whether `typeof getComputedStyle !== 'undefined'`.

src/lib/utils/style-utils.ts
```diff
@@ -0,0 +1,91 @@
...
+ * Find the DOM element's inline style value (if any)
+ */
+export function lookupInlineStyle(element: HTMLElement, styleName: string): string {
+  return getDom().getStyle(element, styleName);
+}
+
+/**
+ * Determine the inline or inherited CSS style
+ */
+export function lookupStyle(element: HTMLElement, styleName: string, inlineOnly = false): any {
```
crisbeto: The return value for this one is a string.

src/lib/utils/testing/dom-tools.ts
```diff
@@ -36,8 +36,14 @@ function getStyle(element: any, stylename: string): string {
   return element.style[stylename];
 }
 
-function hasStyle(element: any, styleName: string, styleValue: string = null): boolean {
-  const value = this.getStyle(element, styleName) || '';
+function hasStyle(element: any,
+                  styleName: string,
+                  styleValue: string = null,
+                  inlineOnly = true): boolean {
+  let searchComputed = () => {
```
crisbeto: Since this is only used once, it seems redundant to define a local function.

src/lib/utils/style-utils.ts
```diff
@@ -0,0 +1,91 @@
+/**
+ * @license
+ * Copyright Google Inc. All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {Renderer2} from '@angular/core';
+import {ɵgetDOM as getDom} from '@angular/platform-browser';
```
ThomasBurleson: Yes, that is intended. Same as in Angular.

src/lib/utils/style-utils.ts
ThomasBurleson: fixed.

src/lib/utils/testing/dom-tools.ts
ThomasBurleson: fixed.

ThomasBurleson:
@crisbeto - Thank you sir for your help and reviews.

PR #386, opened 2017-08-21
fix: update to RendererV2 interface

mhevery:


ThomasBurleson:
@mhevery - my concern here is that merging these changes will `lock` @angular/flex-layout to an unreleased version of Angular. Any suggestions on how to proceed.

mhevery:
I think the best way is to create a `HybridRender` which would be compatible with both new and old interfaces and than cast it to `any`.

```
class HybridRenderer {
  constructor(private renderer: Renderer2) {}

  // Old API needs to be emulated
  setElementClass(el: any, className: string, isAdd: boolean): void { 
    if (isAdd) {
      this.renderer.addClass(el, className);
    } else {
      this.renderer.removeClass(el, className);
    }
  };
  setElementStyle(el: any, styleName: string, styleValue: string): void {
    if (styleValue) {
      this.renderer.setStyle(el, styleName, styleValue);
    } else {
      this.renderer.removeStyle(el, styleName);
    }
  }

  // new API is forwarded
  addClass(el: any, name: string): void { this.renderer.addClass(el, name); }
  removeClass(el: any, name: string): void { this.renderer.removeClass(el, name); }
  setStyle(el: any, style: string, value: any, flags?: RendererStyleFlags2): void { this.render.setStyle(el, style, value); }
  removeStyle(el: any, style: string, flags?: RendererStyleFlags2): void { this.renderer.removeStyle(el, style, flags); }
}
```

you would than:

```
new NgClass(_iterableDiffers, _keyValueDiffers, _ngEl, new HybridRenderer(_renderer));
```

ThomasBurleson:
@misko - love this idea ^; I will make it so before Monday (OoT) this week. Thank you.

ThomasBurleson:
Replaced with #391.

PR #390, opened 2017-08-24
chore(api): repackage API classes

ThomasBurleson:
* repackage API classes to easily distinguish flexbox APIs and  extended responsive APIs
* update testing classes with improved matchers


ThomasBurleson:
@jelbourn - this was reviewed during Beta.9 but reverted from master because it include the `ImgSrcDirective`; which has been removed in this PR. This PR is the basis for the other PRs in the pending queue and should be merged first.

PR #391, opened 2017-08-24
fix(api): use only Renderer2 instances

### LINKED ISSUE #398: ClassDirective throws renderer error with Angular 5.0.0-beta.5
This is because `@angular/flex-layout` doesn't support Angular's `Renderer2` yet, which is the only supported version in Angular `5.0.0-beta.5`. Prior to that release, `Renderer` could be used as well and flex-layout takes advantage of that service in `NgClass` extensions.

screen shot 2017-08-29 at 8 19 11 am

I've opened the original issue over at `@angular/angular` right [HERE](https://github.com/angular/angular/issues/18922)

Also, it seems this will likely be fixed with https://github.com/angular/flex-layout/pull/391 which is pending right now.

ThomasBurleson:
NgClass and NgStyle use Renderer1 instances; which has been deprecated. Instead use Renderer2 and a RendererAdapter.

Thx @mhevery for his solution!

Refs #386. Fixes #398, #408.

### DIFF
--- a/src/lib/api/core/renderer-adapter.ts
+++ b/src/lib/api/core/renderer-adapter.ts
@@ -1,0 +1,74 @@
+import {Renderer2, RendererStyleFlags2} from '@angular/core';
+
+/**
+ * Adapts the 'deprecated' Angular Renderer v1 API to use the new Renderer2 instance
+ * This is required for older versions of NgStyle and NgClass that require
+ * the v1 API (but should use the v2 instances)
+ */
+export class RendererAdapter {
+  constructor(private _renderer: Renderer2) { }
+
+  setElementClass(el: any, className: string, isAdd: boolean): void {
+    if (isAdd) {
+      this._renderer.addClass(el, className);
+    } else {
+      this._renderer.removeClass(el, className);
+    }
+  }
+
+  setElementStyle(el: any, styleName: string, styleValue: string): void {
+    if (styleValue) {
+      this._renderer.setStyle(el, styleName, styleValue);
+    } else {
+      this._renderer.removeStyle(el, styleName);
+    }
+  }
+
+  // new API is forwarded
+  addClass(el: any, name: string): void {
+    this._renderer.addClass(el, name);
+  }
+
+  removeClass(el: any, name: string): void {
+    this._renderer.removeClass(el, name);
+  }
+
+  setStyle(el: any, style: string, value: any, flags?: RendererStyleFlags2): void {
+    this._renderer.setStyle(el, style, value, flags);
+  }
+
+  removeStyle(el: any, style: string, flags?: RendererStyleFlags2): void {
+    this._renderer.removeStyle(el, style, flags);
+  }
+
+  // ******************************************************************
+  // !! Renderer is an abstract class with abstract methods
+  //
+  // These are implementation of those methods... and do NOTHING since
+  // we only use setElementStyle() and setElementClass()
+  // ******************************************************************
+
+  /* tslint:disable */
+  animate()             : any  { throw _notImplemented('animate'); }
+  attachViewAfter()     : void { throw _notImplemented('attachViewAfter'); }
+  detachView()          : void { throw _notImplemented('detachView'); }
+  destroyView()         : void { throw _notImplemented('destroyView'); }
+  createElement()       : any  { throw _notImplemented('createElement'); }
+  createViewRoot()      : any  { throw _notImplemented('createViewRoot'); }
+  createTemplateAnchor(): any  { throw _notImplemented('createTemplateAnchor'); }
+  createText()          : any  { throw _notImplemented('createText'); }
+  invokeElementMethod() : void { throw _notImplemented('invokeElementMethod'); }
+  projectNodes()        : void { throw _notImplemented('projectNodes'); }
+  selectRootElement()   : any  { throw _notImplemented('selectRootElement'); }
+  setBindingDebugInfo() : void { throw _notImplemented('setBindingDebugInfo'); }
+  setElementProperty()  : void { throw _notImplemented('setElementProperty'); }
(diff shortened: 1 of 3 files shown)


ThomasBurleson:
@mhevery - would you review to confirm ?

ThomasBurleson:
@jelbourn - ready for your review also.

ThomasBurleson:

src/lib/api/core/renderer-adapter.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  // ******************************************************************
+  // !! Renderer is an abstract class with abstract methods
+  //
+  // These are implementation of those methods... and do NOTHING since
+  // we only use setElementStyle() and setElementClass()
+  // ******************************************************************
+
+  selectRootElement(): any { }
```
crisbeto: If these methods aren't meant to be used, consider throwing an error or removing the implementation so they fall back to the inherited ones from the `Renderer`?

src/lib/utils/testing/custom-matchers.ts
```diff
@@ -148,76 +155,108 @@ export const customMatchers: jasmine.CustomMatcherFactories = {
...
     };
   }
+
 };
 
 /**
+ * Curried value to function to check styles that are inline or in a stylesheet for the
+ * specified DOM element.
+ */
+function buildCompareStyleFunction(inlineOnly = true) {
```
crisbeto: It seems like these changes are repeated between this PR, #394 and #396. Is this intentional?

src/lib/utils/testing/dom-tools.ts
```diff
@@ -73,6 +76,14 @@ function hasClass(element: any, className: string): boolean {
   return element.classList.contains(className);
 }
 
+function hasAttribute(element: any, attributeName: string): boolean {
+  return element.hasAttribute(attributeName);
```
crisbeto: It doesn't look like most of these DOM utilities are doing much, aside from proxying the native functions with the exact same arguments.

src/lib/api/core/renderer-adapter.ts
ThomasBurleson: We cannot remove the implementation... We want to to appear as a Renderer1 instance but built from a Renderer2. Renderer is an abstract class with many abstract methods. This is we have to implement them EVEN if ignored. 

But we can throw error as not-implemented!

Fixed.

src/lib/utils/testing/custom-matchers.ts
ThomasBurleson: This ^ is because PR 391 will be rebased after #390 gets merged.

jelbourn:
@ThomasBurleson ping me when this one is rebased

src/lib/api/core/renderer-adapter.ts
```diff
@@ -0,0 +1,76 @@
+import {Renderer, Renderer2, RendererStyleFlags2} from '@angular/core';
+
+/**
+ * Adapts the 'deprecated' Angular Renderer v1 API to use the new Renderer2 instance
+ * This is required for older versions of NgStyle and NgClass: which require
+ * the v1 API (but should use the v2 instances)
+ */
+export class RendererAdapter extends Renderer {
```
mhevery: You are still referring to `Renderer`. I think you should drop that.

ThomasBurleson: @mhevery - To manually construct a NgClass I must pass it an instance of a class that `extends Renderer`. Or am I missing something ?

ThomasBurleson: Do you mean change `RendererAdapter` to `HybridAdapter` ?

jelbourn: The whole point of the issue is that the `Renderer` symbol is being removed entirely.

ThomasBurleson: @jelbourn  -If we do this, then we break support for Angular 4.x

```js
// Create an instance NgClass Directive instance only if `ngClass=""` has NOT been defined on
    // the same host element; since the responsive variations may be defined...
    if ( !this._ngClassInstance ) {
      let adapter = new RendererAdapter(_renderer);
      this._ngClassInstance = new NgClass(_iterableDiffers, _keyValueDiffers, _ngEl, adapter);
    }
```

since NgClass in core expects the Renderer.

```js
import { Renderer } from '@angular/core';
export declare class NgClass implements DoCheck {
    private _iterableDiffers;
    private _keyValueDiffers;
    private _ngEl;
    private _renderer;
    private _iterableDiffer;
    private _keyValueDiffer;
    private _initialClasses;
    private _rawClass;
    constructor(_iterableDiffers: IterableDiffers, _keyValueDiffers: KeyValueDiffers, _ngEl: ElementRef, _renderer: Renderer);
```

ThomasBurleson: So I am not sure WHAT to do here because the darn `ngClass` extends an abstract base class instead of implements and interface.

ThomasBurleson: I could just copy the class implementation from ngClass (in core), but that seems to be a clumsy solution. Perhaps that is the only viable solution ?

mhevery: 1) don't implement `Renderer`
2) Make sure that the Shape of the classes matches what `Renderer` expects (this should be noop since you implemented it.)
3) When passing it into `NgClass` just cast it to `any` so that the compiler is happy.

mhevery:
Perfect LGTM

PascalPrecht:
> ping me when this one is rebased

@jelbourn Ping! :)

Would be cool if we could get this in and release as `beta.10` as this is a blocker right now to upgrade from `Angular 5.0.0-beta.4`.

PR #394, opened 2017-08-28
fix(api, class): remove ‘[class]’ selector

### LINKED ISSUE #393: ERROR Error: Uncaught (in promise): TypeError: Cannot read property 'ngDoCheck' of null
I have a form with some flex layout inside. But sometimes when I change the route by a link, the app crash with error

`ERROR Error: Uncaught (in promise): TypeError: Cannot read property 'ngDoCheck' of null`

I found that happen in file flex-layout.es5.js line 2906.  The piece of code is  
```js
ClassDirective.prototype.ngDoCheck = function () {
        if (!this._classAdapter.hasMediaQueryListener) {
            this._configureMQListener();
        }
        this._ngClassInstance.ngDoCheck();
};
```
Then if I patch to 
```js
    ClassDirective.prototype.ngDoCheck = function () {
        if (!this._classAdapter.hasMediaQueryListener) {
            this._configureMQListener();
        }
        if (this._ngClassInstance) {
            this._ngClassInstance.ngDoCheck();
        }
    };
```
the issue doesn't apear anymore.

My context is:
```bash
@angular/cli: 1.3.2
node: 7.10.1
os: linux x64
@angular/animations: 4.3.6
@angular/cdk: 2.0.0-beta.8
@angular/common: 4.3.6
@angular/compiler: 4.3.6
@angular/core: 4.3.6
@angular/flex-layout: 2.0.0-beta.9
@angular/forms: 4.3.6
@angular/http: 4.3.6
@angular/material: 2.0.0-beta.8
@angular/platform-browser: 4.3.6
@angular/platform-browser-dynamic: 4.3.6
@angular/router: 4.3.6
@angular/cli: 1.3.2
@angular/compiler-cli: 4.3.6
@angular/language-service: 4.3.6
```

I dont know if its related, but I have a reactive form and each time the user change something, I save the values. This issue happen after a various number of modifications (and so saves).

```js
    form.valueChanges
      .debounceTime(1000)
      .distinctUntilChanged()
      .filter(v => form.valid)
      .subscribe(v => {
        this.state = FormModelState.Saving;
        (<Observable<Response>>this.save())
          .subscribe(res => {
              this.state = FormModelState.Saved;
              this.error = "";
            },
            e => {
              this.state = FormModelState.SaveError;
              this.error = e;
            }
          );
      });
```
Thanks.

ThomasBurleson:
The host `class` attribute should be considered static and should not be used as a ClassDirective selector. This means that without associated responsive APIs, the ClassDirective will not be instantiated for elements using only the `class` attribute.

* Use the `class` attribute as a responsive fallback value only.
* Improve support of the ngClass directive for string and object assignments.

Fixes #393.

### DIFF
--- a/src/lib/api/core/base-adapter.ts
+++ b/src/lib/api/core/base-adapter.ts
@@ -73,7 +73,9 @@ export class BaseFxDirectiveAdapter extends BaseFxDirective {
     } else if (typeof source === 'string') {
       this._cacheInputString(key, source);
     } else {
-      throw new Error('Invalid class value provided. Did you want to cache the raw value?');
+      throw new Error(
+        `Invalid class value '${key}' provided. Did you want to cache the raw value?`
+      );
     }
   }
 
@@ -123,5 +125,15 @@ export class BaseFxDirectiveAdapter extends BaseFxDirective {
    */
   protected _cacheInputString(key = '', source?: string) {
     this._inputMap[key] = source;
+  }
+
+  /**
+   * Does this directive have 1 or more responsive keys defined
+   * Note: we exclude the 'baseKey' key (which is NOT considered responsive)
+   */
+  public get usesResponsiveAPI() {
+    const totalKeys = Object.keys(this._inputMap).length;
+    const baseValue = this._inputMap[this._baseKey];
+    return  (totalKeys - (!!baseValue ? 1 : 0)) > 0;
   }
 }
--- a/src/lib/api/ext/class.ts
+++ b/src/lib/api/ext/class.ts
@@ -35,7 +35,7 @@ export type NgClassType = string | string[] | Set<string> | {[klass: string]: an
  */
 @Directive({
   selector: `
-    [class], [class.xs], [class.sm], [class.md], [class.lg], [class.xl],
+    [class.xs], [class.sm], [class.md], [class.lg], [class.xl],
     [class.lt-sm], [class.lt-md], [class.lt-lg], [class.lt-xl],
     [class.gt-xs], [class.gt-sm], [class.gt-md], [class.gt-lg],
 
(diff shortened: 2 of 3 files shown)


ThomasBurleson:

src/lib/api/core/base-adapter.ts
```diff
@@ -124,4 +126,17 @@ export class BaseFxDirectiveAdapter extends BaseFxDirective {
   protected _cacheInputString(key = '', source?: string) {
     this._inputMap[key] = source;
   }
+
+  /**
+   * Does this directive have 1 or more responsive keys defined
+   * Note: we exclude the 'baseKey' key (which is NOT considered responsive)
+   */
+  public get usesResponsiveAPI() {
```
crisbeto: Something like this might be more concise:

```ts
public get usesResponsiveAPI() {
  const numKeys = Object.keys(this._inputMap).length;
  return !!(numKeys === 1 || !this._inputMap[this._baseKey]);
}
```

src/lib/api/core/base.ts
```diff
@@ -130,7 +134,7 @@ export abstract class BaseFxDirective implements OnDestroy, OnChanges {
    * and optional restore it when the mediaQueries deactivate
    */
   protected _getDisplayStyle(source?: HTMLElement): string {
-    let element: HTMLElement = source || this._elementRef.nativeElement;
+    let element: HTMLElement = source || this.nativeElement;
```
crisbeto: Consider switching the method signature to `source = this.nativeElement` here and in `_applyStyleToElement`.

src/lib/api/ext/class.ts
```diff
@@ -77,41 +77,21 @@ export class ClassDirective extends BaseFxDirective
...
-   * Initial value of the `class` attribute; used as
-   * fallback and will be merged with nay `ngClass` values
-   */
-  get initialClasses() : string {
-    return  this._classAdapter.queryInput('_rawClass') || "";
-  }
+  @Input('class.gt-xs')   set classGtXs(val: NgClassType) { this._classAdapter.cacheInput('classGtXs', val); }
+  @Input('class.gt-sm')   set classGtSm(val: NgClassType) { this._classAdapter.cacheInput('classGtSm', val); }
+  @Input('class.gt-md')   set classGtMd(val: NgClassType) { this._classAdapter.cacheInput('classGtMd', val); }
+  @Input('class.gt-lg')   set classGtLg(val: NgClassType) { this._classAdapter.cacheInput('classGtLg', val); }
```
crisbeto: All of these setters will generate a lot of ES5 JS. Have you considered using `ngOnChanges`  and looping through the changed inputs to cache them instead?

src/lib/api/flexbox/layout-wrap.ts
```diff
@@ -142,7 +142,7 @@ export class LayoutWrapDirective extends BaseFxDirective implements OnInit, OnCh
   }
 
   protected get flowDirection(): string {
-    let computeFlowDirection = () => this._getFlowDirection(this._elementRef.nativeElement);
+    let computeFlowDirection = () => this._getFlowDirection(this.nativeElement);
     return this._layoutWatcher ? this._layout : computeFlowDirection();
```
crisbeto: Defining a function just to call it immediately seems a little wasteful. Consider returning `this._getFlowDirection(this.nativeElement)` instead?

src/lib/api/ext/class.ts
naveedahmed1: @crisbeto was this addressed before merge?

src/lib/api/core/base-adapter.ts
ThomasBurleson: @crisbeto - that logic is not correct since we are checking for 1 or more non-baseKey values.

```js
public get usesResponsiveAPI() {
  const numKeys = Object.keys(this._inputMap).length;
  return ( !this._inputMap[this._baseKey] && (numKeys > 0));
}
```

src/lib/api/ext/class.ts
ThomasBurleson: @crisbeto - there is a PR for this ^ feature in @angular:  https://github.com/angular/angular/issues/13355

src/lib/api/core/base-adapter.ts
ThomasBurleson: Fixed.

PR #396, opened 2017-08-29
fix(ObservableMedia): properly announce 'xs' activation at startup

### LINKED ISSUE #388: subscribing to ObservableMedia doesnt start with value on load of XS mq
Open the example below with a XS mobile size screen  and ObservableMedia doesn't fire on page load, but it seems to work fine for all other media query sizes.

screen shot 2017-08-28 at 5 38 13 pm

https://angular-wfwpzp.stackblitz.io/

source:
https://stackblitz.com/edit/angular-wfwpzp?file=app%2Fapp.component.ts

ThomasBurleson:
MatchMedia was not properly announcing XS activations for applications with startup viewports < 600px.
Improve the specificity of the 'xs' media query enables proper MatchMedia notifications at startup.

Fixes #388.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -10,7 +10,7 @@ copies of the Software, and to permit persons to whom the Software is
 furnished to do so, subject to the following conditions:
 
 The above copyright notice and this permission notice shall be included in
-all copies or substantial portions of the  Software.
+all copies or substantial portions of the Software.
 
 THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
 IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
--- a/src/lib/api/core/base.ts
+++ b/src/lib/api/core/base.ts
@@ -133,9 +133,8 @@ export abstract class BaseFxDirective implements OnDestroy, OnChanges {
    * Note: this allows use to preserve the original style
    * and optional restore it when the mediaQueries deactivate
    */
-  protected _getDisplayStyle(source?: HTMLElement): string {
-    let element: HTMLElement = source || this.nativeElement;
-    return lookupStyle(element, 'display');
+  protected _getDisplayStyle(source: HTMLElement = this.nativeElement): string {
+    return lookupStyle(source || this.nativeElement, 'display');
   }
 
   /**
@@ -164,7 +163,7 @@ export abstract class BaseFxDirective implements OnDestroy, OnChanges {
    */
   protected _applyStyleToElement(style: StyleDefinition,
                                  value?: string | number,
-                                 nativeElement?: any) {
+                                 nativeElement: any = this.nativeElement) {
     let element = nativeElement || this.nativeElement;
     applyStyleToElement(this._renderer, element, style, value);
   }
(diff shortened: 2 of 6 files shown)


ThomasBurleson:

LICENSE
```diff
@@ -10,7 +10,7 @@ copies of the Software, and to permit persons to whom the Software is
 furnished to do so, subject to the following conditions:
 
 The above copyright notice and this permission notice shall be included in
-all copies or substantial portions of the Software.
+all copies or substantial portions of the  Software.
```
crisbeto: Seems like you added an extra space in here.

src/lib/api/core/base.ts
```diff
@@ -130,7 +134,7 @@ export abstract class BaseFxDirective implements OnDestroy, OnChanges {
    * and optional restore it when the mediaQueries deactivate
    */
   protected _getDisplayStyle(source?: HTMLElement): string {
-    let element: HTMLElement = source || this._elementRef.nativeElement;
+    let element: HTMLElement = source || this.nativeElement;
```
crisbeto: Instead of doing `element = source || this.nativeElement`, you can change the method signature to be `protected _getDisplayStyle(source: = this.nativeElement): string`.

src/lib/api/core/base.ts
```diff
@@ -161,7 +165,7 @@ export abstract class BaseFxDirective implements OnDestroy, OnChanges {
   protected _applyStyleToElement(style: StyleDefinition,
                                  value?: string | number,
                                  nativeElement?: any) {
-    let element = nativeElement || this._elementRef.nativeElement;
+    let element = nativeElement || this.nativeElement;
```
crisbeto: Same approach as https://github.com/angular/flex-layout/pull/396/files#diff-172881d6558fa5a20f715647057c40c2R137.

src/lib/utils/testing/custom-matchers.ts
```diff
@@ -148,76 +155,108 @@ export const customMatchers: jasmine.CustomMatcherFactories = {
...
+      allPassed = allPassed && elHasStyle;
+      if (!elHasStyle) {
+        extendObject(found, current);
+      }
+    });
+
+    return {
+      pass: allPassed,
+      get message() {
+        const expectedValueStr = (typeof styles === 'string') ? styles : JSON.stringify(styles);
```
crisbeto: To make the output of this a little more readable, you can pass in the indentation level to `JSON.stringify`:  `JSON.stringify(styles, null, 2)`.

src/lib/api/core/base.ts
ThomasBurleson: I did not know you could initialize with a dynamic property:

```js
protected _getDisplayStyle(source: HTMLElement = this.nativeElement): string {
```

so cool.

ThomasBurleson: @crisbeto - note sure if you default value *solves* all the scenarios.
My solution will account for calls like `_getDisplayStyle(null)` where your solution will fail with that.

src/lib/api/core/base.ts
ThomasBurleson: See comment above regarding why `let element = nativeElement || this.nativeElement` is better than a default parameter value.

src/lib/utils/testing/custom-matchers.ts
ThomasBurleson: fixed.

src/lib/api/core/base.ts
crisbeto: Makes sense, didn't know that scenario had to be handled.

ThomasBurleson:
@crisbeto - I just added a unit tests for XS startup activation announcements.

crisbeto:
Seems like you have a linting error @ThomasBurleson.

PR #418, opened 2017-09-13
fix(api): check platform before accessing nativeElement

cupidi:
Style, Class and ImgSrc now check if the platform is browser before accessing nativeElement properties.

Fixes #408

Issue #408, opened 2017-09-06
StyleDirective throws renderer error '_ngEl.nativeElement.getAttribute is not a function'


aaronte:
Trying to run `@angular/flex-layout#2.0.0-beta.9` with universal and getting this error:

```bash
Unhandled Promise rejection: _ngEl.nativeElement.getAttribute is not a function ; Zone: <root> ; Task: Promise.then ; Value: { TypeError: _ngEl.nativeElement.getAttribute is not a function
    at new StyleDirective (/user_code/node_modules/@angular/flex-layout/bundles/flex-layout.umd.js:3087:61)
    at createClass (/user_code/node_modules/@angular/core/bundles/core.umd.js:10970:26)
    at createDirectiveInstance (/user_code/node_modules/@angular/core/bundles/core.umd.js:10788:37)
    at createViewNodes (/user_code/node_modules/@angular/core/bundles/core.umd.js:12229:49)
    at callViewAction (/user_code/node_modules/@angular/core/bundles/core.umd.js:12675:13)
    at execComponentViewsAction (/user_code/node_modules/@angular/core/bundles/core.umd.js:12584:13)
    at createViewNodes (/user_code/node_modules/@angular/core/bundles/core.umd.js:12256:5)
    at createRootView (/user_code/node_modules/@angular/core/bundles/core.umd.js:12124:5)
    at Object.createProdRootView [as createRootView] (/user_code/node_modules/@angular/core/bundles/core.umd.js:12809:12)
    at ComponentFactory_.create (/user_code/node_modules/@angular/core/bundles/core.umd.js:9896:46)
  __zone_symbol__currentTask: 
   ZoneTask {
     _zone: 
      Zone {
        _properties: {},
        _parent: null,
        _name: '<root>',
        _zoneDelegate: [Object] },
     runCount: 0,
     _zoneDelegates: null,
     _state: 'notScheduled',
     type: 'microTask',
     source: 'Promise.then',
     data: undefined,
     scheduleFn: undefined,
     cancelFn: null,
     callback: [Function],
     invoke: [Function] } } TypeError: _ngEl.nativeElement.getAttribute is not a function
    at new StyleDirective (/user_code/node_modules/@angular/flex-layout/bundles/flex-layout.umd.js:3087:61)
    at createClass (/user_code/node_modules/@angular/core/bundles/core.umd.js:10970:26)
    at createDirectiveInstance (/user_code/node_modules/@angular/core/bundles/core.umd.js:10788:37)
[...]

Issue #273, opened 2017-05-01
ngStyle conflict with ngStyle.alias


alexfung888:
Supposedly they are independent

```
<div
  [ngStyle]="{'transform': slideStyleString, 'transition': 'transform 0.75s ease-in-out'}"
  [ngStyle.xs]="{'color': 'green'}"
>TESTING</div>
```
However, if you can the value of `slideStyleString` above, the color will be lost.

Problem exists in 2.0.0-beta.7 and 2.0.0-beta.8.
Problem not present in 2.0.0-rc.1

[See plunker. HTML is in src/app/test-app.ts](https://plnkr.co/edit/AnvTM3rL473EMqBqKWHY?p=preview)
The first DIV stays in color. Click the slide out button and the second DIV loses color and turns black.

alexfung888:
An even simply example:
```
<div
  [ngStyle]="{'fontSize': sizeString}"
  [ngStyle.xs]="{'color': 'green'}"
  [ngStyle.sm]="{'color': 'red'}"
  [ngStyle.md]="{'color': 'blue'}"
  [ngStyle.lg]="{'color': 'purple'}"
  [ngStyle.xl]="{'color': 'pink'}"
>TESTING</div>
<button (click)='sizeString="20pt"'>Size 20</button>
<button (click)='sizeString="10pt"'>Size 10</button>   
```

ThomasBurleson:
@alexfung888 - While the [ngStyle Wiki](https://github.com/angular/flex-layout/wiki/ngStyle-API#merging-styles) says styles will be merged. Let me confirm again Angular core expectations.

M-Adoo:
@ThomasBurleson  Any progress on this one ?

mackelito:
@ThomasBurleson will this be fixed in the upcoming release?

ThomasBurleson:
no. This fix will be available in Beta.10 (or new RC1 if we can overwrite the existing invalid one).

ThomasBurleson:
The latest nightly build provides fixes for ngStyle

![screen shot 2017-09-15 at 5 54 22 pm](https://user-images.githubusercontent.com/210413/30506125-f8c4b6e2-9a3e-11e7-8b47-de6b5c98da43.png)

```html
<div
        [ngStyle]="{'color': 'red', 'fontSize': sizeString}"
        [ngStyle.xs]="{'color': 'green'}"
        [ngStyle.sm]="{'color': 'red', 'fontSize': '12pt'}"
[...]

PR #419, opened 2017-09-13
fix(api, class, style): remove deprecated selectors

### LINKED ISSUE #410: Error "Can't bind to 'ngStyle.md' since it isn't a known property of 'div'"
Error at runtime.

Should there be a [ngStyle.md] selector in style.ts? There is one for [style.md].

ThomasBurleson:
* all `[style]`, `[style.xxx]`,`[class]`, and `[class.xxx]` selectors have been removed
  * only `ngStyle` and `ngClass` selectors support responsive suffices
  * now use of raw `style` attribute will NOT instantiate a StyleDirective
  * now use of raw `class` attribute will NOT instantiate a ClassDirective
* The `ClassDirective` now simply adds responsive enhancements to the core `NgClass` directive
* The `StyleDirective` now simply adds responsive enhancements to the core `NgStyle` directive
* Added missing `[ngStyle.md]` selector
* Added `isBrowser()` universal checks for `getAttribute()` and `getStyle()` queries

Fixes #410, #408, #273, #420. Closes #418.

BREAKING CHANGES

* **ngStyle**, **ngClass** should be used for responsive selectors and changes. Raw `style` and `class` attributes no longer support responsive suffices.

##### Before

```html
<div  fxLayout
  [class.xs]="['xs-1', 'xs-2']"
  [style]="{'font-size': '10px', 'margin-left' : '13px'}"
  [style.xs]="{'font-size': '16px'}"
  [style.md]="{'font-size': '12px'}">
</div>
```

##### Now

```html
<div  fxLayout
  [ngClass.xs]="['xs-1', 'xs-2']"
  [ngStyle]="{'font-size': '10px', 'margin-left' : '13px'}"
  [ngStyle.xs]="{'font-size': '16px'}"
  [ngStyle.md]="{'font-size': '12px'}">
</div>
```


ThomasBurleson:
@mmalerba - ready for your review and g3 testing.

src/lib/api/core/base-adapter.ts
```diff
@@ -20,6 +20,16 @@ import {MediaMonitor} from '../../media-query/media-monitor';
 export class BaseFxDirectiveAdapter extends BaseFxDirective {
 
   /**
+   * Does this directive have 1 or more responsive keys defined
+   * Note: we exclude the 'baseKey' key (which is NOT considered responsive)
+   */
+  public hasResponsiveAPI() {
+    const totalKeys = Object.keys(this._inputMap).length;
+    const baseValue = this._inputMap[this._baseKey];
+    return  (totalKeys - (!!baseValue ? 1 : 0)) > 0;
```
mmalerba: can this just be `return this.hasResponsiveAPI(this._baseKey)`?

ThomasBurleson: Sorry @mmalerba - I do not understand your question here ^.

ThomasBurleson: the `_baseKey` is the only non-responsive key. We want to check if 1 or more responsive keys are defined/in-use.

mmalerba: I'm just saying that this class extends a class with a 1-parameter method of the same name, it looks like this no-parameter version could just call the 1-parameter version

ThomasBurleson: Fixed.

src/lib/api/ext/class.ts
```diff
@@ -154,51 +124,50 @@ export class ClassDirective extends BaseFxDirective
...
-   *  Identify the activated input value and update the ngClass iterables...
-   *  needs ngDoCheck() to actually apply the values to the element
+   * Initial lookup of raw 'class' value (if any)
    */
-  protected _updateNgClass(value?: NgClassType) {
-    if (this._ngClassAdapter.mqActivation) {
-      value = this._ngClassAdapter.mqActivation.activatedInput;
+  protected _fallbackToKlass() {
+    if (!this._base.queryInput('ngClass')) {
+      this.ngClassBase = this._getAttributeValue('class');
```
ThomasBurleson: Premise: When a breakpoint deactivates, the fallback value in the non-responsive input will be used.
But what happens when the fallback/default input is not defined?

Consider this logic here in `_fallbackToKlass()`. Is this logic valid?

This methods effectively says that if a responsive breakpoint deactivates (e.g. `[ngClass.md]`) and the default `[ngClass]` has not been defined, then we should use the `class` attribute value (if any) as the fallback value.

The challenge is the any `ngClass` responsive input could be a string, array, or map. Each do different things.

1) Is this a valid assumption to fallback/restore the initial `class` value?

PR #444, opened 2017-10-02
fix(fxLayoutGap): update fxLayoutGap when only 1 child is remaining

### LINKED ISSUE #433: fxLayoutGap not being updated after removing element from DOM
[Plunker](https://plnkr.co/edit/mwwofp51BtJf401Xa8dd?p=preview)
[Gif](http://recordit.co/uNHDjl4w1X)

I have some components that are utilizing `fxHide`/`fxShow`; when these components are made visible, the `fxLayoutGap` isn't being correctly applied, as you can see in the GIF/Plunker. Looking at the HTML, I can see that the `margin-right` property isn't being applied to the newly-visible component, and only happens to a component that was previously hidden at first, e.g. in the Plunker, you should notice that the desktop tile will still retain its `margin-right` property even after it is hidden, whilst the mobile-only tile will never have it.

aldo-roman:
fixes #433

### DIFF
--- a/src/lib/api/flexbox/layout-gap.spec.ts
+++ b/src/lib/api/flexbox/layout-gap.spec.ts
@@ -139,6 +139,35 @@ describe('layout-gap directive', () => {
       });
 
     }));
+    
+    it('should add update gap styles when only 1 row is remaining', async(() => {
+      let template = `
+              <div fxLayoutAlign='center center' fxLayoutGap='13px'>
+                  <div fxFlex *ngFor='let row of rows'></div>
+              </div>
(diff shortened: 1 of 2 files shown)


aldo-roman:
@ThomasBurleson any comment on this PR? Thanks!

ThomasBurleson:
@aldo-roman - This is a good fix. Can you provide one unit test to verify the fix in this PR ?

aldo-roman:
Hi @ThomasBurleson ! Done! Sorry for not including them in the first place.

PR #448, opened 2017-10-04
chore: update to use material mat- prefix

kara:
Angular Material modules no longer have the Md- prefix (now Mat-).


andrewseguin:
LGTM =D

PR #466, opened 2017-10-22
fix(universal): correct fxLayout rendering, update LICENSE

### LINKED ISSUE #423: bug(fxLayout, universal): layout direction is wrong
When using Universal rendering, fxLayout gets picked up, but wrongly applied as `flex-direction=row` instead of `flex-direction=column`.

Confirmed that this only happens on the latest commit, which means a wire got crossed somewhere in the refactoring for NgClass and NgStyle (e461d17471f713deaf4663e0a0438822f8f0ad86).

CaerusKaru:
* Fixes #423
* Remove default value of fxLayout to avoid confusion
* Removed browser checks from style-utils since getDOM will always
  return an implementation with the methods implemented
* Added note about getComputedStyle PR pending
* Updated RxJS to 5.5.0 and added necessary fixes,
   updated other deps
* Added better null and type checking throughout
* Updated LICENSE, see https://github.com/angular/material2/pull/7589


ThomasBurleson:
@CaerusKaru - has this PR been rebased with latest in `master` ?

CaerusKaru:
@ThomasBurleson it’s all synced up

ThomasBurleson:
@CaerusKaru - Ok, I will review. Thank you for this PR.

ThomasBurleson:
@CaerusKaru - why did you use a `replaySubject(1)` instead of a `behaviorSubject`?

CaerusKaru:
@ThomasBurleson `BehaviorSubject` gives an incorrect assumption that all `fxLayout`s start with a `row` value. While it's true that `fxLayout` defaults to row (and still does), anything that subscribes to this value later down the chain is given a false impression.

Enter `ReplaySubject(1)`, which is exactly like the `BehaviorSubject`, except without the erroneous "default" value. Now if something subscribes, it will wait for the first real value, and anything that subscribes later will still get the current value.

CaerusKaru:
This is not a dealbreaker for me, and I'd be happy to change it back since it doesn't appear to affect functionality. I just thought it was an improved alternative.

ThomasBurleson:
@CaerusKaru - Thank you for this PR. Merged in SHA 8dcae026

PR #469, opened 2017-10-24
chore: use a for loop instead Array.map() when the return value is discarded.

bowenni:
We're enabling a check in google's internal TypeScript compiler that detects calls to well-known functions and methods that return a value that is discarded. So this pattern will become a compilation error soon.


bowenni:
Ping :)

bowenni:
Ping

ThomasBurleson:
@bowenni - Thanks, this will be in Beta.13.

PR #474, opened 2017-10-31
build: import individual rx operators

CaerusKaru:
* Removes 'rxjs/operators' import in favor of individual imports,
  adds lint rule to prevent that type of import
* Maintains parity with angular/material2#8160

PR #521, opened 2017-11-30
chore: remove deprecated fxLayoutWrap

CaerusKaru:


CaerusKaru:
Same error as #474

CaerusKaru:
Blocked by #523

ThomasBurleson:
@CaerusKaru - seems reasonable to remove verbose API when option is already provided in `fxLayout`.

ThomasBurleson:
@CaerusKaru - while I like this. Perhaps we should deprecate this feature current HEAD and the next release and then formally remove it in two releases ?

CaerusKaru:
From the comment in `fxLayoutWrap`, it looks like it's been deprecated at least since beta.10. But I think it could wait until we bump major to 5.0.0

ThomasBurleson:
Yes. Let's keep this PR pending until then.

PR #527, opened 2017-12-07
fix: apply correct RTL margins

### LINKED ISSUE #274: RTL support?
I'm not sure if I'm configuring this incorrect of that this library doesn't support RTL content.
The following is my layout html:
```html
<div fxLayout="row" fxLayoutAlign="space-between none">
        <div fxFlex="20">
            <md-slide-toggle [(ngModel)]="advanced"></md-slide-toggle>
        </div>
        <div fxFlex="60">
            <span (click)="advanced = !advanced"><i class="fa icon-gear"></i> {{resources.toggleAdvancedSettings}}</span>
        </div>
        <div fxFlex>
            <div fxLayout="col" fxLayoutAlign="start end">
                <span class="cursor-pointer" (click)="closeSidebar()"><i class="fa fa-lg icon-close"></i></span>
            </div>
        </div>
    </div>
```
What I want to accomplish is close to the following:
image
Only the "X" in the image should be aligned to the end, which in RTL case is left instead of right.
I'm using angular 4 with `[dir]="rtl"` directive so I was hoping this is in sync with angular's RTL support...

atscott:
Fixes #274

### DIFF
--- a/src/lib/api/flexbox/layout-gap.ts
+++ b/src/lib/api/flexbox/layout-gap.ts
@@ -24,6 +24,7 @@ import {Subscription} from 'rxjs/Subscription';
 
 import {BaseFxDirective} from '../core/base';
 import {LayoutDirective} from './layout';
+import {Directionality} from '../../bidi/directionality';
 import {MediaChange} from '../../media-query/media-change';
 import {MediaMonitor} from '../../media-query/media-monitor';
 import {LAYOUT_VALUES} from '../../utils/layout-validator';
@@ -45,6 +46,7 @@ export class LayoutGapDirective extends BaseFxDirective implements AfterContentI
   protected _layout = 'row';  // default flex-direction
   protected _layoutWatcher: Subscription;
   protected _observer: MutationObserver;
+  private _directionWatcher: Subscription;
 
   /* tslint:disable */
  @Input('fxLayoutGap')       set gap(val) { this._cacheInput('gap', val); }
@@ -70,12 +72,15 @@ export class LayoutGapDirective extends BaseFxDirective implements AfterContentI
               renderer: Renderer2,
               @Optional() @Self() container: LayoutDirective,
               private _zone: NgZone,
-              @Inject(PLATFORM_ID) platformId: Object) {
+              @Inject(PLATFORM_ID) platformId: Object,
+              private _directionality: Directionality) {
     super(monitor, elRef, renderer, platformId);
 
     if (container) {  // Subscribe to layout direction changes
       this._layoutWatcher = container.layout$.subscribe(this._onLayoutChange.bind(this));
     }
+    this._directionWatcher =
+        this._directionality.change.subscribe(this._updateWithValue.bind(this));
   }
 
   // *********************************************
@@ -107,6 +112,9 @@ export class LayoutGapDirective extends BaseFxDirective implements AfterContentI
     }
     if (this._observer) {
       this._observer.disconnect();
+    }
+    if (this._directionWatcher) {
+      this._directionWatcher.unsubscribe();
     }
   }
 
@@ -196,7 +204,7 @@ export class LayoutGapDirective extends BaseFxDirective implements AfterContentI
       case 'row' :
       case 'row-reverse':
       default :
-        key = 'margin-right';
+        key = this._directionality.value === 'rtl' ? 'margin-left' : 'margin-right';
         break;
     }
     margins[key] = value;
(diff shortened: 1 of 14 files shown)


CaerusKaru:
The stance is that it is very much encouraged. angular/cdk is a dependency of this library (which may end up getting merged into Angular Material anyway). Please feel free to use it instead of forcing us to maintain two sets of the same code.

src/lib/api/directionality.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * The directionality (LTR / RTL) context for the application (or a subtree of it).
+ * Exposes the current direction and a stream of direction changes.
+ */
+@Injectable()
+export class Directionality {
+  /** The current 'ltr' or 'rtl' value. */
+  readonly value: Direction = 'ltr';
+
+  constructor(@Optional() @Inject(DIR_DOCUMENT) _document?: any) {
```
CaerusKaru: Is there a reason you're not importing the `DOCUMENT` token from `angular/common`? This is why Travis is failing on prerender, it doesn't have access to the global browser `document` object

atscott:
The @angular/cdk was only in dev dependencies (for the demo, I think). I've moved it to the dependency list and am now using the Directionality injectable from the bidi module.

atscott:
FYI - current broken test is also failing in master

CaerusKaru:
Rebase with master should fix broken test

atscott:
Done, thanks.

atscott:
Friendly post-holiday ping. Anything that needs to be done to get this merged? RTL support is pretty important to my team.

CaerusKaru:
It would be great if you could add a demo that displays LTR in the demo-app, just something simple that documents the capability.

Also, please change your commit message to something in line with the Angular contribution guidelines, found [here](https://github.com/angular/angular/blob/master/CONTRIBUTING.md#-commit-message-guidelines)

Finally, the PR lists this as a partial fix for #274, are there cases that this does not cover, and can you list that in the commit message?

Other than that the PR LGTM, I'll see what I can do to get it merged in the next week or so (it has to go through internal presubmit tests first)

atscott:
Updated commit message (I think). 

The "partial fix" in the description was from when I wasn't using the angular cdk dir directive. This should be a full fix of what was requested in that bug.

I haven't looked much yet at where to put this in the demo app - I wasn't able to get it to run correctly.

CaerusKaru:
You need to squash your commits into one with a rebase. If there is an issue with the demo-app, can you post it as a new issue so we can take a look?

atscott:
Rebased and filed #547

CaerusKaru:
#547 has been patched in #549. Please incorporate the change (it's one line) and add a demo to the demo-app, leaving out the one-line change in your final commit.

atscott:
Added the demo and updated the implementations to subscribe to the direction changes.

src/lib/api/flexbox/layout-gap.ts
```diff
@@ -67,12 +69,16 @@ export class LayoutGapDirective extends BaseFxDirective implements AfterContentI
...
+              private _zone: NgZone,
+              private readonly _directionality: Directionality) {
     super(monitor, elRef, renderer);
 
     if (container) {  // Subscribe to layout direction changes
       this._layoutWatcher = container.layout$.subscribe(this._onLayoutChange.bind(this));
     }
+
+    this._directionWatcher = 
+        this._directionality.change.subscribe(() => this._updateWithValue());
```
CaerusKaru: This should be the same as the _layoutWatcher subscription: 

```typescript
this._directionality.change.subscribe(this._updateWithValue.bind(this));
```

CaerusKaru:
Demo looks great, thanks 👍

CaerusKaru:
@atscott Last thing: lint is failing because of trailing whitespace. Just remove and recommit

atscott:
Thanks for the review and guidance!

ThomasBurleson:
@atscott - this is good work. Thank you.
Currently Flex-Layout can be used standalone. Your additions will add an explicit dependency on the Angular CDK. I am reluctant to add that dependency for only the Directionality.

CaerusKaru:
@ThomasBurleson The PR originally duplicated the CDK directionality code, would you want to incorporate that or a similar patch instead of adding the direct dependency?

IMO adding the CDK as a dependency seems reasonable, especially since a lot of people using Flex Layout likely are already using the CDK with Material (having started with layout using AngularJS Material), but I understand your reservations about adding more dependencies than necessary.

CaerusKaru:
@atscott After discussion, we're going to postpone this at least until the next major release. We want to make sure Flex Layout is more stable before we include the CDK

ThomasBurleson:
@atscott - The reason we want to clone the Directionality class is to avoid introducing an explicit dependency on the CDK at this time. 

> Later, major portions of the library [ e.g.  `/src/lib/mediaquery`] may be migrated to the CDK and then the @angular/flex-layout dependency on the CDK will be acceptable.

atscott:
Alright, that sounds good. I made those changes (had trouble rebasing so kind of redid the whole thing) and I think everything's all set.

PR #539, opened 2017-12-19
fix(fxLayout): add support for inline-flex

### LINKED ISSUE #525: feature: support inline-flex similar to inline-block
Feature request, or proposal:
Why is **inline-flex** not supported?

ThomasBurleson:
Flexbox CSS supports `display:inline-flex` to indicate that the flexbox container should be inline; similar to the `display:inline-block`.
When used with the **fxLayout** API, this fix will support the **inline** option:

*  `fxLayout="row inline"`
*  `fxLayout="column wrap inline"`

Fixes #525.

### DIFF
--- a/src/lib/api/flexbox/layout.spec.ts
+++ b/src/lib/api/flexbox/layout.spec.ts
@@ -141,6 +141,71 @@ describe('layout directive', () => {
 
   });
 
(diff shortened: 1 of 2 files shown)


danwellman:
The documentation should be updated to include this

PR #563, opened 2018-01-15
refactor: remove private Angular APIs

### LINKED ISSUE #402: Remove usage of private angular core API
Current code uses imports and code such as:

```js
import {ɵgetDOM as getDom} from '@angular/platform-browser';

return getDom().getStyle(element, styleName);
```

This is a private API. Replace this will solution similar to that used in @angular/material.

CaerusKaru:
* Remove private API references to getDom()
* Add PLATFORM_ID to all components

Closes #553.
Fixes #402.


ThomasBurleson:
Closes (and merged with SHA 703add02ad4a30467b84bb6f3aa8358d4e01364d)

PR #566, opened 2018-01-15
chore: update to Angular v5.2 and update peer dep requirements

### LINKED ISSUE #559: support for angular 5.2.x
It say the followings when 5.2.x or higher installed:
@angular/flex-layout@2.0.0-beta.12 requires a peer of @angular/core@~5.1.0 but none was installed.
@angular/flex-layout@2.0.0-beta.12 requires a peer of @angular/common@~5.1.0 but none was installed.

It shouldnt say the followings when 5.2.x or higher installed:
@angular/flex-layout@2.0.0-beta.12 requires a peer of @angular/core@~5.1.0 but none was installed.
@angular/flex-layout@2.0.0-beta.12 requires a peer of @angular/common@~5.1.0 but none was installed.

install @angular/core@ and @angular/common@ with higher version than 5.1.x e.g: 5.2.x

To be up to date

5.2.x

CaerusKaru:
* Update to Angular v5.2 and TypeScript v2.6
* Update build config to conform with angular/material2#9355
* Update travis to node 8 and npm latest
* Bring package build patterns in line with angular/material2
* Enable strictNullChecks and refactor as necessary

Succeeds PR #560

Fixes #559


DevVersion:
Would it be possible to also update the `publish-build-artifacts.sh` script? There has been a critical fix to the publish script, that should be fixed on flex-layout as well (to guarantee a less flakier CI)

* See: https://github.com/angular/material2/blob/master/scripts/deploy/publish-build-artifacts.sh

CaerusKaru:
Looks like @ThomasBurleson beat me to it with fe877c9

PR #567, opened 2018-01-15
feat(ssr): enhance support for Universal and SSR with stylesheets

### LINKED ISSUE #373: Universal pre-rendered styles are not correct
With last nightly build, there are no more blocking errors with Flex Layout in Universal, but the pre-rendered styles are not correct.

Repro : 
- `git clone https://github.com/cyrilletuzi/nguniversal-expressengine-httpinterceptor.git`
- `git checkout -b flexlayoutissue`
- `npm install`
- `npm start`
- go to localhost:3000 : rendered correctly with JavaScript (vertical in small screens, horizontal in desktop screens with 3 stretched items by row)
- disable JavaScript to test what the server is pre-rendering and refresh : not rendered correctly (horizontal with the 4 items on the same line, on all screen sizes)

The only flex relative code is in `src/app/app.component` (the other files are just Universal boilerplate, and `innerHTML` is just to quickly test different heights) : 

```html
<div fxLayout.xs="column" fxLayoutWrap.xs="nowrap">
  <div class="test" fxFlex.gt-xs="calc(100% / 3)" *ngFor="let item of list" [innerHTML]="item"></div>
</div>
```

##### Mobile Viewport 

screen shot 2017-08-09 at 10 53 16 am

##### Desktop Viewport

screen shot 2017-08-09 at 10 24 07 am

----

##### Mobile Viewport 

screen shot 2017-08-09 at 10 53 01 am

##### Desktop Viewport

screen shot 2017-08-09 at 10 29 35 am

CaerusKaru:
* Add `StyleService` class to manage application and retrieval of styles from elements in a platform-agnostic manner
* Add virtual stylesheet to store server styles, which applies default styles when breakpoint overrides are not present
* While not in the browser (ssr), intercept all style calls and reroute them to the virtual stylesheet
* For server-side rendering, add a new type of MediaQueryList similar to the MockMediaQueryList to support manual activation/deactivation of breakpoints
* Add jasmine testing mode for SSR
* Add FlexLayoutServerModule to invoke SSR styling
* Remove unnecessary Renderer references and replace them with DOM APIs
* Add whitespace debugging mode for server styles

Fixes #373.

> See [Design Doc](https://docs.google.com/document/d/1fg04ihw42dJJHGd6fugdiBe39iJot8aErhiE7CjwfmQ/edit#)


CaerusKaru:
Example output for the following HTML markup:

```html
<div class="night-theme">
  <div fxLayout="row" fxLayout.xs="column" style="height:500px" ngxSplit="row">
    <div fxFlex="30%" ngxSplitArea class="c1r1">
      <div class="c1r1_header">Column #1 - Row #1</div>
      <ul>
        <li>2 Columns: 30% + 70%</li>
        <li>2nd Column: 2 rows</li>
        <li>2nd Column: 50% + 50%</li>
      </ul>
    </div>
    <div class="handle handle-row" ngxSplitHandle>
      <i class="material-icons">&#xE25D;</i>
    </div>
    <div fxFlex.xs="70%" fxFlex.gt-md="50%" fxFlex.lg="60%" ngxSplitArea>
      <div fxLayout="column" fxFlexFill ngxSplit="column">
        <div fxFlex="50%" ngxSplitArea class="c2r1_body">
          <div class="c2r1_header">Column #2 - Row #1</div>
          <h1>Layout Dashboard</h1>
          <p>
            Demonstrate use of ngxSplit with the Flex-Layout API
            and flexbox css layouts.
            <br/><br/>
            Haxx0r ipsum cd ctl-c Starcraft concurrently salt unix baz class bar linux
            January 1, 1970 syn for mutex daemon todo mountain dew recursively. Mainframe
            wannabee machine code hack the mainframe do void python bin big-endian break
            tcp ddos emacs public frack.Over clock headers data private *.* pwned
            fork script kiddies.
          </p>
        </div>
        <div class="handle handle-column" ngxSplitHandle>
          <i class="material-icons">&#xE25D;</i>
        </div>
        <div fxFlex="50%" ngxSplitArea class="c2r2">
          <div class="c2r2_header">Column #2 - Row #2</div>
          <ul>
            <li>List Item #1</li>
            <li>List Item #2</li>
            <li>List Item #3</li>
          </ul>
        </div>
      </div>
    </div>
  </div>
</div>
```

```css
@media all {
     .flex-layout-0 {
         height: 100%;
         min-height: 100%;
         min-width: 100%;
         width: 100%;
         -webkit-flex-direction: column;
         box-sizing: border-box;
         display: flex;
         flex-direction: column;
    }
     .flex-layout-1 {
         -webkit-flex-direction: row;
         box-sizing: border-box;
         display: flex;
         flex-direction: row;
    }
     .flex-layout-2 {
         -webkit-flex: 1 1 0.000000001px;
         box-sizing: border-box;
         flex: 1 1 0.000000001px;
         max-width: 30%;
         -webkit-flex-basis: 100%;
         -webkit-flex-grow: 1;
         -webkit-flex-shrink: 1;
         flex-basis: 100%;
         flex-grow: 1;
         flex-shrink: 1;
    }
     .flex-layout-3 {
         -webkit-flex: 1 1 0.000000001px;
         box-sizing: border-box;
         flex: 1 1 0.000000001px;
    }
     .flex-layout-4 {
         -webkit-flex: 1 1 0.000000001px;
         box-sizing: border-box;
         flex: 1 1 0.000000001px;
         max-height: 50%;
         -webkit-flex-basis: 100%;
         -webkit-flex-grow: 1;
         -webkit-flex-shrink: 1;
         flex-basis: 100%;
         flex-grow: 1;
         flex-shrink: 1;
    }
     .flex-layout-5 {
         -webkit-flex: 1 1 0.000000001px;
         box-sizing: border-box;
         flex: 1 1 0.000000001px;
         max-height: 50%;
         -webkit-flex-basis: 100%;
         -webkit-flex-grow: 1;
         -webkit-flex-shrink: 1;
         flex-basis: 100%;
         flex-grow: 1;
         flex-shrink: 1;
    }
}
 @media (min-width: 1280px) and (max-width: 1919px) {
     .flex-layout-1 {
         -webkit-flex-direction: row;
         box-sizing: border-box;
         display: flex;
         flex-direction: row;
    }
     .flex-layout-3 {
         -webkit-flex-basis: 100%;
         -webkit-flex-grow: 1;
         -webkit-flex-shrink: 1;
         box-sizing: border-box;
         flex-basis: 100%;
         flex-grow: 1;
         flex-shrink: 1;
         max-width: 60%;
    }
}
 @media (max-width: 1279px) {
     .flex-layout-1 {
         -webkit-flex-direction: row;
         box-sizing: border-box;
         display: flex;
         flex-direction: row;
    }
     .flex-layout-3 {
         -webkit-flex: 1 1 0.000000001px;
         box-sizing: border-box;
         flex: 1 1 0.000000001px;
    }
}
 @media (min-width: 1280px) {
     .flex-layout-1 {
         -webkit-flex-direction: row;
         box-sizing: border-box;
         display: flex;
         flex-direction: row;
    }
     .flex-layout-3 {
         -webkit-flex-basis: 100%;
         -webkit-flex-grow: 1;
         -webkit-flex-shrink: 1;
         box-sizing: border-box;
         flex-basis: 100%;
         flex-grow: 1;
         flex-shrink: 1;
         max-width: 50%;
    }
}
 @media (min-width: 960px) and (max-width: 1279px) {
     .flex-layout-1 {
         -webkit-flex-direction: row;
         box-sizing: border-box;
         display: flex;
         flex-direction: row;
    }
     .flex-layout-3 {
         -webkit-flex: 1 1 0.000000001px;
         box-sizing: border-box;
         flex: 1 1 0.000000001px;
    }
}
 @media (min-width: 0px) and (max-width: 599px) {
     .flex-layout-1 {
         -webkit-flex-direction: column;
         box-sizing: border-box;
         display: flex;
         flex-direction: column;
    }
     .flex-layout-2 {
         -webkit-flex-basis: 100%;
         -webkit-flex-grow: 1;
         -webkit-flex-shrink: 1;
         box-sizing: border-box;
         flex-basis: 100%;
         flex-grow: 1;
         flex-shrink: 1;
         max-height: 30%;
    }
     .flex-layout-3 {
         -webkit-flex: 1 1 0.000000001px;
         box-sizing: border-box;
         flex: 1 1 0.000000001px;
         max-height: 70%;
         -webkit-flex-basis: 100%;
         -webkit-flex-grow: 1;
         -webkit-flex-shrink: 1;
         flex-basis: 100%;
         flex-grow: 1;
         flex-shrink: 1;
    }
}
```

CaerusKaru:
Note: this needs a testing mode where we can change the PLATFORM_ID flag from browser mode to server mode and confirm all tests still pass.

naveedahmed1:
@CaerusKaru should I create a new issue to track the combining of classes for the common styles?

CaerusKaru:
@naveedahmed1 You can add it to #373

CaerusKaru:
@jelbourn @vikerman

src/lib/api/ext/img-src.ts
```diff
@@ -57,12 +60,16 @@ export class ImgSrcDirective extends BaseFxDirective implements OnInit, OnChange
...
-    super(monitor, elRef, renderer, platformId);
-    this._cacheInput('src', elRef.nativeElement.getAttribute('src') || '');
+  constructor(protected _elRef: ElementRef,
+              protected _monitor: MediaMonitor,
+              protected _styler: StyleService,
+              @Inject(PLATFORM_ID) protected _platformId: Object,
+              @Optional() @Inject(SERVER_TOKEN) protected _serverModuleLoaded: boolean) {
+    super(_monitor, _elRef, _styler);
+    this._cacheInput('src', _elRef.nativeElement.getAttribute('src') || '');
+    if (isPlatformServer(this._platformId) && this._serverModuleLoaded) {
```
vikerman: Can you explain a bit on how the serverModuleLoaded flag is used?

vikerman:
The platform-server part and associated tests looks good to me(I didn't go through the actual styles logic). 

Just need a small clarification.

(In the future it would help if the PRs are collection of smaller PRs/commits)

src/lib/api/ext/img-src.ts
CaerusKaru: The flag is to determine if the `FlexLayoutServerModule` is included (since it's the only module that provides that token).

Loading the server module is optional -- if a user doesn't load it, it means they want the inline styles as a fallback (it also prevents this from being a breaking change, since it's totally opt-in).

vikerman: Will this get into the docs?

CaerusKaru: Absolutely. Docs will be in a follow-up PR once this has all been finalized

src/lib/media-query/match-media.ts
```diff
@@ -5,42 +5,83 @@
...
+    if (this._listeners.indexOf(listener) === -1) {
+      this._listeners.push(listener);
+    }
+    if (this._isActive) {
+      listener(this);
+    }
+  }
+
+  removeListener(_: MediaQueryListListener) {
+  }
```
jelbourn: Add comment in function body here like
```ts
// Don't need to remove listeners in the server environment.
```

src/lib/media-query/match-media.ts
```diff
@@ -5,42 +5,83 @@
...
- * EventHandler callback with the mediaQuery [range] activates or deactivates
+ * Special server-only class to simulate a MediaQueryList and
+ * - supports manual activation to simulate mediaQuery matching
+ * - manages listeners
  */
-export interface MediaQueryListListener {
-  // Function with Window's MediaQueryList argument
-  (mql: MediaQueryList): void;
-}
+export class ServerMediaQueryList implements MediaQueryList {
```
jelbourn: This should be in a separate file
(generally all server-specific stuff should be in separate files so that you can better ensure you're only getting what you need on the client)

src/lib/media-query/match-media.ts
```diff
@@ -51,19 +92,36 @@ export interface MediaQueryList {
  */
 @Injectable()
 export class MatchMedia {
-  protected _registry: Map<string, MediaQueryList>;
+  protected _registry: Map<string, MediaQueryList|ServerMediaQueryList>;
```
jelbourn: Instead of using a union type, it would be better to use an interface and an `InjectionToken`. When using a union type, you may end up getting the server-side code on the client

src/lib/server/server-provider.ts
```diff
@@ -0,0 +1,117 @@
...
+
+let UNIQUE_CLASS = 0;
+const DEBUG_FLAG = false;
+
+/**
+ * create @media queries based on a virtual stylesheet
+ * * Adds a unique class to each element and stores it
+ *   in a shared classMap for later reuse
+ */
+function formatStyle(stylesheet: Map<HTMLElement, Map<string, string|number>>,
```
jelbourn: Something like `generateCss`?

src/lib/server/server-provider.ts
```diff
@@ -0,0 +1,117 @@
...
+                     _document: Document,
+                     mediaQuery: string,
+                     classMap: Map<HTMLElement, string>) {
+  let styleText = DEBUG_FLAG ? `
```
jelbourn: Rather than formatting the css manually, it would probably be better to use some existing node css formatter / prettifier (still only when debugging). As it is now, the method is pretty hard to mentally parse.

src/lib/server/server-provider.ts
```diff
@@ -0,0 +1,117 @@
...
+  BreakPoint,
+  BREAKPOINTS,
+  CLASS_NAME,
+  MatchMedia,
+  ServerStylesheet,
+  SERVER_TOKEN,
+} from '@angular/flex-layout';
+
+let UNIQUE_CLASS = 0;
+const DEBUG_FLAG = false;
```
jelbourn: `IS_DEBUG_MODE`?

src/lib/server/server-provider.ts
jelbourn: Add JsDoc parameter descriptions

src/lib/server/server-provider.ts
```diff
@@ -0,0 +1,117 @@
...
+ * platform
+ * format the static @media queries for all breakpoints
+ * to be used on the server and append them to the <head>
+ */
+export function addStyles(serverSheet: ServerStylesheet,
+                          matchMedia: MatchMedia,
+                          _document: Document,
+                          breakpoints: BreakPoint[]) {
+  // necessary because of angular/angular/issues/14485
+  const res = () => {
```
jelbourn: `res` isn't a very meaningful variable name

src/lib/server/server-provider.ts
```diff
@@ -0,0 +1,117 @@
...
+  return styleText;
+}
+
+/**
+ * Add or remove static styles depending on the current
+ * platform
+ * format the static @media queries for all breakpoints
+ * to be used on the server and append them to the <head>
+ */
+export function addStyles(serverSheet: ServerStylesheet,
```
jelbourn: Add JsDoc param descriptions

src/lib/server/server-provider.ts
```diff
@@ -0,0 +1,117 @@
...
+                          matchMedia: MatchMedia,
+                          _document: Document,
+                          breakpoints: BreakPoint[]) {
+  // necessary because of angular/angular/issues/14485
```
jelbourn: I don't see how this issue is related

src/lib/server/server-provider.ts
```diff
@@ -0,0 +1,117 @@
...
+  const res = () => {
+    const styleTag = _document.createElement('style');
+    const classMap = new Map<HTMLElement, string>();
+    const defaultStyles = new Map(serverSheet.stylesheet);
```
jelbourn: Add descriptions for these variables?

src/lib/server/server-provider.ts
```diff
@@ -0,0 +1,117 @@
...
+    let styleText = formatStyle(defaultStyles, _document, 'all', classMap);
+
+    breakpoints.reverse();
```
jelbourn: Add comment for what this block of code is doing?

src/lib/server/server-provider.ts
```diff
@@ -0,0 +1,117 @@
...
+  return res;
+}
+
+/**
+ *  Provider to set static styles on the server
+ */
+export const SERVER_PROVIDERS = [
+  {
+    provide: BEFORE_APP_SERIALIZED,
+    useFactory: addStyles,
```
jelbourn: I would create a separate thing like `FLEX_SSR_SERIALIZER_FACTORY` that then calls the more specific functions

src/lib/server/server-provider.ts
jelbourn: `addStyles` is not very specific. Could the name be something more like `generateStaticFlexLayoutStyles`? It would also be good if the description went into more detail about how it works.

src/lib/utils/styling/browser-provider.ts
```diff
@@ -0,0 +1,42 @@
...
+ */
+import {
+  APP_BOOTSTRAP_LISTENER,
+  PLATFORM_ID,
+  InjectionToken,    // tslint:disable-line:no-unused-variable
+  ComponentRef,      // tslint:disable-line:no-unused-variable
+} from '@angular/core';
+import {DOCUMENT, isPlatformBrowser} from '@angular/common';
+
+export function removeStyles(_document: Document, platformId: Object) {
```
jelbourn: Description?

src/lib/utils/styling/browser-provider.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {
+  APP_BOOTSTRAP_LISTENER,
+  PLATFORM_ID,
+  InjectionToken,    // tslint:disable-line:no-unused-variable
+  ComponentRef,      // tslint:disable-line:no-unused-variable
```
jelbourn: Why do these unused symbols need to stay here?

src/lib/utils/styling/browser-provider.ts
```diff
@@ -0,0 +1,42 @@
...
+} from '@angular/core';
+import {DOCUMENT, isPlatformBrowser} from '@angular/common';
+
+export function removeStyles(_document: Document, platformId: Object) {
+  // necessary because of angular/angular/issues/14485
```
jelbourn: How is this issue related?

src/lib/utils/styling/server-stylesheet.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {Injectable} from '@angular/core';
+
+@Injectable()
+export class ServerStylesheet {
```
jelbourn: Class and method descriptions?

src/lib/utils/styling/server-token.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {InjectionToken} from '@angular/core';
+
+export const SERVER_TOKEN = new InjectionToken<boolean>('FlexLayoutServerLoaded');
```
jelbourn: Description?

src/lib/utils/styling/styler.ts
```diff
@@ -0,0 +1,128 @@
...
+ */
+import {Inject, Injectable, Optional, PLATFORM_ID} from '@angular/core';
+import {isPlatformBrowser, isPlatformServer} from '@angular/common';
+
+import {applyCssPrefixes} from '../auto-prefixer';
+import {ServerStylesheet} from './server-stylesheet';
+import {SERVER_TOKEN} from './server-token';
+
+@Injectable()
+export class StyleService {
```
jelbourn: Class description?

On naming: my rule is to avoid the term "Service" in class names because it doesn't tell you anything about what the class actually does. You'll notice in material we have things like `ScrollDispatcher`, `FocusMonitor`, `InteractivityChecker`, etc. instead of `ScrollService`, `FocusService`, `InteracitivityService`. Adding a more specific noun makes the name much more meaningful.

src/lib/utils/styling/styler.ts
```diff
@@ -0,0 +1,128 @@
...
+  /**
+   * Find the DOM element's inline style value (if any)
+   */
+  lookupInlineStyle(element: HTMLElement, styleName: string): string {
+    return element.style[styleName] || element.style.getPropertyValue(styleName);
+  }
+
+  /**
+   * Determine the inline or inherited CSS style
+   * @TODO(CaerusKaru): platform-server has no implementation for getComputedStyle
```
jelbourn: What's the action on this TODO? The whole idea of `getComputedStyle` inherently can't work on the server (without being driven by an entire browser rendering engine)

src/lib/utils/styling/styler.ts
```diff
@@ -0,0 +1,128 @@
...
+    let value = '';
+    if (element) {
+      let immediateValue = value = this.lookupInlineStyle(element, styleName);
+      if (!immediateValue) {
+        if (isPlatformBrowser(this._platformId)) {
+          if (!inlineOnly) {
+            value = getComputedStyle(element).getPropertyValue(styleName);
+          }
+        } else {
+          if (this._serverModuleLoaded) {
```
jelbourn: There are a lot of places throughout the PR where you're explicitly performing a check for browser vs. server and performing a different action for each. In general it would be better to delegate to some injectable with a common interface that uses a different implementation on each platform.

src/lib/utils/styling/styler.ts
```diff
@@ -0,0 +1,128 @@
...
+      }
+    });
+  }
+}
+
+/**
+ * Definition of a css style. Either a property name (e.g. "flex-basis") or an object
+ * map of property name and value (e.g. {display: 'none', flex-order: 5})
+ */
+export type StyleDefinition = string | { [property: string]: string | number | null };
```
jelbourn: Why do either an object literal or string instead of just always doing an object literal?

src/lib/utils/styling/browser-provider.ts
CaerusKaru: Because otherwise we get the following TS error: 

```
error TS4023: Exported variable 'MATCH_MEDIA_PROVIDER' has or is using name 'InjectionToken' from external module "/Users/main/flex-layout/node_modules/@angular/core/src/di/injection_token" but cannot be named
```

src/lib/utils/testing/custom-matchers.ts
```diff
@@ -204,7 +205,7 @@ export const customMatchers: jasmine.CustomMatcherFactories = {
  * specified DOM element.
  */
 function buildCompareStyleFunction(inlineOnly = true) {
-  return function (actual: any, styles: { [k: string]: string } | string) {
+  return function (actual: any, styles: { [k: string]: string } | string, styler: StyleService) {
```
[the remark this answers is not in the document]
CaerusKaru: Discussed offline, unfortunately very necessary :-1:

src/lib/media-query/match-media.ts
CaerusKaru: Added

src/lib/media-query/match-media.ts
CaerusKaru: Added to a separate file with a separate implementation of `MatchMedia` for the server as suggested below

src/lib/media-query/match-media.ts
CaerusKaru: Added, great suggestion!

src/lib/server/server-provider.ts
CaerusKaru: Fixed

src/lib/server/server-provider.ts
CaerusKaru: Do you have a CSS formatter in mind? Maybe this can be done in a follow-up PR?

src/lib/server/server-provider.ts
CaerusKaru: Added

src/lib/server/server-provider.ts
CaerusKaru: Removed

src/lib/server/server-provider.ts
CaerusKaru: Added

src/lib/utils/styling/browser-provider.ts
CaerusKaru: Removed

src/lib/utils/styling/styler.ts
CaerusKaru: This is legacy code, removed. (The idea was that when calling a function, you could call it with string and value individually instead of providing an object)

src/lib/utils/styling/styler.ts
CaerusKaru: Now that the `MatchMedia` has been changed (see above), the only three places this still happens are the following: ImgSrc, ShowHide, and Styler. In each case, the check only changes about two lines of code in very critical ways. I think it would duplicate a lot of code to change and as it stands now doesn't benefit much to factor out.

src/lib/utils/styling/styler.ts
CaerusKaru: Changed to `NOTE`

src/lib/utils/styling/server-token.ts
CaerusKaru: Added

src/lib/utils/styling/server-stylesheet.ts
CaerusKaru: Added

src/lib/utils/styling/styler.ts
CaerusKaru: I vaguely recall @ThomasBurleson saying the same. The original name was `StyleUtils`, I'll change it back.

src/lib/utils/styling/browser-provider.ts
CaerusKaru: Added

src/lib/server/server-provider.ts
CaerusKaru: Removed from above comment

src/lib/server/server-provider.ts
CaerusKaru: Added

src/lib/server/server-provider.ts
CaerusKaru: Added

src/lib/server/server-provider.ts
CaerusKaru: Added

CaerusKaru: Added

src/lib/server/server-provider.ts
CaerusKaru: Added

CaerusKaru:
@jelbourn Comments addressed, PTAL

src/lib/media-query/match-media-provider.ts
```diff
@@ -0,0 +1,39 @@
...
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {
+  NgZone,
+  PLATFORM_ID,
+  InjectionToken,    // tslint:disable-line:no-unused-variable
```
jelbourn: You can avoid suppressing the lint error by explicitly using `InjectionToken` in the place where it is inferred

src/lib/media-query/match-media-provider.ts
```diff
@@ -0,0 +1,39 @@
...
+import {DOCUMENT, isPlatformServer} from '@angular/common';
+
+import {MatchMedia} from './match-media';
+import {ServerMatchMedia} from './server-match-media';
+
+/**
+ * Create a version of MatchMedia compatible with the current
+ * platform
+ */
+export function MEDIA_QUERY_LIST_FACTORY(platformId: Object,
```
jelbourn: Instead of having one provider that returns either service, there should just be two different providers. The appropriate provider is then included in each of the browser / server module.

src/lib/media-query/server-match-media.ts
```diff
@@ -0,0 +1,136 @@
...
+import {MatchMedia} from './match-media';
+
+/**
+ * Special server-only class to simulate a MediaQueryList and
+ * - supports manual activation to simulate mediaQuery matching
+ * - manages listeners
+ */
+export class ServerMediaQueryList implements MediaQueryList {
+  private _isActive = false;
+  private _listeners: Array<MediaQueryListListener> = [];
```
jelbourn: `MediaQueryListListener[]`

src/lib/media-query/server-match-media.ts
```diff
@@ -0,0 +1,136 @@
...
+   * listeners and clearing the internal list
+   */
+  destroy() {
+    this.deactivate();
+    this._listeners = [];
+  }
+
+  /**
+   * Notify all listeners that 'matches === TRUE'
+   */
```
jelbourn: You can collapse short comment to a single line:
```ts
/** Notify all listeners that 'matches === TRUE' */
```

src/lib/server/server-provider.ts
```diff
@@ -0,0 +1,140 @@
...
+  BreakPoint,
+  BREAKPOINTS,
+  CLASS_NAME,
+  MatchMedia,
+  ServerStylesheet,
+  SERVER_TOKEN,
+  ServerMatchMedia
+} from '@angular/flex-layout';
+
+let UNIQUE_CLASS = 0;
```
jelbourn: `UNIQUE_CLASS` shouldn't be caps case because it's not constant. I typically call this `nextId` in material ([example](https://github.com/angular/material2/blob/3aceb7361cc34ad987f7b1ca39339d3203248341/src/cdk/accordion/accordion-item.ts#L24))

src/lib/utils/styling/browser-provider.ts
jelbourn: You can avoid the suppression by explicitly using them in the code where they are now inferred.

src/lib/media-query/match-media-provider.ts
CaerusKaru: Fixed

src/lib/media-query/server-match-media.ts
CaerusKaru: Fixed (and fixed the source for this)

src/lib/media-query/server-match-media.ts
CaerusKaru: Fixed

src/lib/server/server-provider.ts
CaerusKaru: Fixed

CaerusKaru:
Comments addressed again, PTAL

ThomasBurleson:
Merged with SHA cf5266adc71cb377a54020d64a500d5f789450ed

PR #576, opened 2018-01-18
revert: revert typescript to 2.5.x as ng 5.2 not compatible with ts 2.6.x yet

akkumar:
Artifact of #560.

ng 5.2.x is not compatible with typescript 2.6.x (yet) . 

It is blocked on angular/angular#21591 . 

More details here at: https://github.com/angular/devkit/pull/394 . 

Given that - may be we should let the typescript version to be 2.5.x as before. 

Cc @ThomasBurleson @CaerusKaru

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5206,7 +5206,7 @@
         "resolve-dependency-path": "1.0.2",
         "sass-lookup": "1.1.0",
         "stylus-lookup": "1.0.2",
-        "typescript": "2.6.2"
+        "typescript": "2.5.3"
       }
     },
     "fill-range": {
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@akkumar - how is ng 5.2 **not** compatible with TS 2.6 ?

CaerusKaru:
The TypeScript team is aware of this issue and has triaged it for TS 2.8. Since the blocking issue is not a runtime/build issue, I see no reason to revert back to TS 2.5 for this library. Installing TS 2.6 for Flex Layout is a) not global, and b) Flex specific (ie it doesn't require you to install TS 2.6 to install the library, just to build it locally from source and contribute).

If this is an issue for you contributing locally, I suggest you downgrade on an individual basis, since it looks like this only affects VS Code for now.

If Angular Core reverts to TS 2.5.x, then we can take another look at this issue, but the Core and the CLI (they have  [PR](https://github.com/angular/angular-cli/pull/9085) in flight) have moved on, and so have we.

Closing this for now, will reopen if Angular reverts or if this becomes a build issue for Flex Layout.

PR #581, opened 2018-01-19
build: set ci sudo to false

CaerusKaru:


CaerusKaru:
This is to bring parity with angular/angular#21641

@DevVersion can you review this please?

test/karma.conf.js
```diff
@@ -86,7 +86,7 @@ module.exports = (config) => {
     browserDisconnectTimeout: 20000,
     browserNoActivityTimeout: 240000,
     captureTimeout: 120000,
-    browsers: ['Chrome_1024x768'],
+    browsers: process.env.TRAVIS ? ['ChromeHeadlessCI'] : ['ChromeHeadlessCISandbox'],
```
DevVersion: This shouldn't be necessary. This never runs in Travis

CaerusKaru: Fixed

PR #584, opened 2018-01-20
chore(versions): upgrade dependency-tree to 6.0.0

akkumar:
Address the following warning. 

```console
npm WARN typescript-eslint-parser@1.0.2 requires a peer of typescript@~2.0.3 but none is installed. You must install peer dependencies yourself.
```
by upgrading dependency-tree to 6.0.0 (madge depends on the same. Independently, pr logged with madge as well to avoid explicitly upgrading dependency-tree - https://github.com/pahen/madge/pull/147 ).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3956,15 +3956,32 @@
       "dev": true
     },
     "dependency-tree": {
-      "version": "5.11.0",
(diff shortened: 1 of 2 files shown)


CaerusKaru:
Since you have a PR open for `madge`, I see no reason to add an extra dev dependency here, even temporarily. It's only a warning, after all.

PR #586, opened 2018-01-22
feat(MediaObserver): add MediaObserver to replace ObservableMedia

### LINKED ISSUE #297: feat(ObservableMedia): extend Observable
I know it's a little ugly to have non-operator methods on `Observable`, but I think the `ObservableMedia` API would be a little nicer if it were an actual RxJS `Observable` instead of just implementing `Subscribable`. Maybe @benlesh has stronger feelings about not mixing Observables with other types :).

To support this would basically just require implementing `lift` method, and extending `Observable`. Happy to submit a PR.

CaerusKaru:
* Add MediaObservable class that returns an Observable instead of creating a Subscribable
* Deprecate ObservableMedia and MediaService since they are no longer needed

BREAKING CHANGE:
* `ObservableMedia` is deprecated. Please use `MediaObserver`, which has compatible APIs

Fixes #297

### DIFF
--- a/src/lib/core/breakpoints/break-points-provider.ts
+++ b/src/lib/core/breakpoints/break-points-provider.ts
@@ -44,7 +44,7 @@ export function buildMergedBreakPoints(_custom?: BreakPoint[],
   }, options || {});
 
   return () => {
-    // Order so the defaults are loaded last; so ObservableMedia will report these last!
+    // Order so the defaults are loaded last; so MediaObserver will report these last!
     let defaults = (options && options.orientations) ?
       ORIENTATION_BREAKPOINTS.concat(DEFAULT_BREAKPOINTS) : DEFAULT_BREAKPOINTS;
 
--- a/src/lib/core/match-media/match-media.spec.ts
+++ b/src/lib/core/match-media/match-media.spec.ts
@@ -5,8 +5,6 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {Observable} from 'rxjs/Observable';
-
 import {TestBed, inject, async} from '@angular/core/testing';
 
 import {MediaChange} from '../media-change';
@@ -15,8 +13,7 @@ import {MockMatchMedia} from './mock/mock-match-media';
 import {DEFAULT_BREAKPOINTS_PROVIDER} from '../breakpoints/break-points-provider';
 import {BreakPointRegistry} from '../breakpoints/break-point-registry';
 import {MatchMedia} from './match-media';
-import {ObservableMedia} from '../observable-media/observable-media';
-import {OBSERVABLE_MEDIA_PROVIDER} from '../observable-media/observable-media-provider';
+import {MediaObserver} from '../media-observer/media-observer';
 
 describe('match-media', () => {
   let matchMedia: MockMatchMedia;
@@ -27,7 +24,7 @@ describe('match-media', () => {
       providers: [
         BreakPointRegistry,           // Registry of known/used BreakPoint(s)
         DEFAULT_BREAKPOINTS_PROVIDER, // Supports developer overrides of list of known breakpoints
-        {provide: MatchMedia, useClass: MockMatchMedia}
+        {provide: MatchMedia, useClass: MockMatchMedia},
       ]
     });
   });
@@ -124,7 +121,7 @@ describe('match-media', () => {
 describe('match-media-observable', () => {
   let breakPoints: BreakPointRegistry;
   let matchMedia: MockMatchMedia;
-  let mediaQuery$: Observable<MediaChange>;
+  let mediaQuery$: MediaObserver;
 
   beforeEach(() => {
     // Configure testbed to prepare services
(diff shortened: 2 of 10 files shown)


CaerusKaru:
@benlesh can you please review this?

benlesh:
The code looks solid. I'm not sure about the architecture or the use case. 

Generally I advise against subclassing Observable these days. It's a little like subclassing a function. It might be better to have a class that had a method returning the Observable, I dunno. Probably worth some design discussion and a document. Without that I'm really speculating. I don't have a lot of context in this part of Angular.

I think you've done a good job with what you've set out to do, from the looks of it. Observable extended properly. You may or may not want to override lift, to get that custom method to carry. But that won't work well with TypeScript anyhow.

CaerusKaru:
@benlesh Following your suggestion, I've changed `MediaObservable` to something that has a function that returns the source `Observable` instead of extending it. Thank you very much for your feedback!

src/lib/media-query/media-observable.ts
```diff
@@ -0,0 +1,95 @@
...
+import {filter} from 'rxjs/operators/filter';
+
+import {MediaChange} from './media-change';
+import {BreakPointRegistry} from './breakpoints/break-point-registry';
+import {MatchMedia} from './match-media';
+import {BreakPoint} from './breakpoints/break-point';
+import {mergeAlias} from '../utils/add-alias';
+
+@Injectable()
+export class MediaObservable {
```
benlesh: Now the name is probably a misnomer. It's no longer an "Observable". Maybe MediaSource? MediaWatcher? I don't know.

src/lib/media-query/media-observable.ts
```diff
@@ -0,0 +1,95 @@
...
+
+  /**
+   * Test if specified query/alias is active.
+   */
+  isActive(alias): boolean {
+    let query = this._toMediaQuery(alias);
+    return this.matchMedia.isActive(query);
+  }
+
+  observe(): Observable<MediaChange> {
```
benlesh: The name could be more specific, and you might consider whether or not you want this to be a property. Since Observables do nothing until you execute them, they are more like functions than Promises are. So a method that returns an Observable is more like a higher order function, where a property that returns an Observable is more like a method. If that makes sense.

I'm not saying that this fits this use case, it's just something I see.

src/lib/media-query/media-observable.ts
```diff
@@ -0,0 +1,95 @@
...
+    const activationsOnly = (change: MediaChange) => {
+      return change.matches === true;
+    };
+    const addAliasInformation = (change: MediaChange) => {
+      return mergeAlias(change, this._findByQuery(change.mediaQuery));
+    };
+    const excludeOverlaps = (change: MediaChange) => {
+      let bp = this.breakpoints.findByQuery(change.mediaQuery);
```
benlesh: this appears to be the same as `_findByQuery` used above (and defined below).

src/lib/media-query/media-observable.ts
```diff
@@ -0,0 +1,95 @@
...
+      return !bp ? true : !(this.filterOverlaps && bp.overlapping);
```
benlesh: const noBreakpoints = !bp;
const noOverlaps = !(this.filterOverlaps && bp.overlapping);
return noBreakpoints || noOverlaps;

Or just

return !bp || !(this.filterOverlaps && bp.overlapping);

... if you're trying to code golf this for some reason.

src/lib/media-query/media-observable.ts
```diff
@@ -0,0 +1,95 @@
...
+    };
+
+    return this.matchMedia.observe()
```
benlesh: the observe() method here is going to get the same comment as I made above... should this be a property? (I'm not saying it *should*, I'm asking... because usually that's the case)

CaerusKaru: In the case of `MatchMedia`, `observe` takes an optional string argument

src/lib/media-query/media-observable.ts
CaerusKaru: Excellent, added

src/lib/media-query/media-observable.ts
CaerusKaru: For the case of this service, I think it makes sense, but I'll talk it over with @ThomasBurleson to see if we will ever need parameters for this like with `MatchMedia`

src/lib/media-query/media-observable.ts
CaerusKaru: In keeping with Material, I think we're going to go with `MediaObserver`

src/lib/media-query/media-observable.ts
CaerusKaru: Great catch, fixed

sandangel:
Hello, is this MediaObserver overlapped with `@angular/cdk/layout` BreakpointObserver?

CaerusKaru:
@sandangel It is not a direct overlap, but rather one more compatible with Flex Layout's design. The CDK layout engine is a more general utility.

CaerusKaru:
@ThomasBurleson Addressed comments and ready for review :+1:

ardatan:
Why don't you use cdk providers?

CaerusKaru:
@ardatan As mentioned above, our use case is not 1:1 with the CDK

CaerusKaru:
Marking this as on hold while we consider switching to the `@angular/cdk` media engine

CaerusKaru:
Closing in favor of moving to the CDK's media engine

PR #597, opened 2018-02-05
chore(docs): replace image HTML with Markdown syntax

CaerusKaru:
* Replace HTML in the docs with Markdown
* Add description to links


ThomasBurleson:
@CaerusKaru - let's change the commit message to `chore(docs): replace image HTML with Markdown syntax`.

PR #610, opened 2018-02-11
build: fix runtime errors not being logged out on karma initialization

CaerusKaru:
* Equivalent commit to angular/material#9888


CaerusKaru:
@DevVersion Can you take a look at this?

PR #612, opened 2018-02-12
build: don't skip master build when only md files have changed

CaerusKaru:
* Equivalent commit to angular/material2#9879


CaerusKaru:
@DevVersion Can you please take a look at this?

PR #618, opened 2018-02-15
build: conditional travis build stages

CaerusKaru:
* Equivalent commit to angular/material2#9949

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,14 +21,15 @@ jobs:
     - env: "MODE=lint"
     - env: "MODE=aot"
     - env: "MODE=prerender"
-    # Closure Compiler CI check is temporarily disabled until a new version of
-    # the tool is released with https://github.com/google/closure-compiler/pull/2600
-    # - env: "MODE=closure-compiler"
     - env: "MODE=saucelabs_required"
     - env: "MODE=browserstack_required"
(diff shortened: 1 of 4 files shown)


CaerusKaru:
@DevVersion can you take a look at this?

.travis.yml
```diff
@@ -21,14 +21,14 @@ jobs:
...
     - env: "MODE=aot"
     - env: "MODE=prerender"
-    # Closure Compiler CI check is temporarily disabled until a new version of
-    # the tool is released with https://github.com/google/closure-compiler/pull/2600
-    # - env: "MODE=closure-compiler"
     - env: "MODE=saucelabs_required"
     - env: "MODE=browserstack_required"
     - env: "MODE=travis_required"
     - stage: Deploy
       env: "DEPLOY_MODE=build-artifacts"
```
DevVersion: Actually the `if` statements are missing here. So it's not really about conditional build stages. On Material we want to push build artifacts not in another build stage, because it can happen that BrowserStack/Saucelabs time out too often.

CaerusKaru: Good point, I had the formatting wrong and thought the `if` statements came before the `env`. I added a check for the `publish-artifacts` stage, can you PTAL again?

PR #627, opened 2018-02-22
refactor: outsource the CSS prettier for debugging server styles

CaerusKaru:
* Remove custom stylings when debugging server styles and replace
  it with a CSS prettier

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,5 @@ npm-debug.log
 testem.log
 /.chrome
 !/test/karma.conf.js
+
+src/apps/universal-app/package-lock.json
--- a/package-lock.json
+++ b/package-lock.json
@@ -1703,8 +1703,7 @@
     "abbrev": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/abbrev/-/abbrev-1.1.1.tgz",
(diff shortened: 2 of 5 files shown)


ThomasBurleson:
@CaerusKaru - I do not want to bundle `beautify` as a dependency for flex-layout.
I will close this PR. BTW, here is the code refactor I wanted:

```ts
function generateCss(stylesheet: Map<HTMLElement, Map<string, string|number>>,
                     mediaQuery: string,
                     classMap: Map<HTMLElement, string>) {
  let css = '';
  stylesheet.forEach((styles, el) => {
    let keyVals   = '';
    let className = getClassName(el,  classMap);

    styles.forEach((v, k) => {
        keyVals += v ? formatOutput(`${k}:${v};`) : '';
    });

    // Build list of CSS styles; each with a className
    css += formatOutput(`.${className} {`) + keyVals + formatOutput('}');
  });

  // Group 1 or more styles (each with className) in a specific mediaQuery
  return formatOutput(`@media ${mediaQuery} {`) + css + formatOutput('}');
}

/**
 * Prefix css segment with linefeed for easy
 * debugging purposes.
 */
function formatOutput(css:string):string {
  return IS_DEBUG_MODE ? '\n' + css : css;
}

/**
 * Get className associated with CSS styling
 * If not found, generate global className and set
 * association.
 */
function  getClassName(stylesheet, classMap) {
  let className = classMap.get(stylesheet);
   if (!className) {
     className = `${CLASS_NAME}${nextId++}`;
     classMap.set(stylesheet, className);
   }
   stylesheet.classList.add(className);

   return className;
}
```

PR #628, opened 2018-02-25
fix(fxFlex): apply correct flex-basis stylings

### LINKED ISSUE #277: Setting flex-basis to 0 results in max-width: 0
I'm trying to set up rows with equal-width columns. I can tell flexbox to not consider the contents in its calculations by setting a flex-basis of 0. This lets flex-grow distribute all available space in consistent proportions.

I tried the following:

```
<div fxLayout="row">
    <div fxFlex="1 1 0px">Grows to 25% width</div>
    <div fxFlex="3 1 0px">Grows to 75% width</div>
</div>
```

What resulted was both child divs having `max-width: 0px` set, which ruined my layout. If I remove this max-width property, the row renders correctly. [Here's a plunker demonstrating the issue](https://plnkr.co/edit/LKpWisJKoxM6UZlMqVsY?p=preview).

### LINKED ISSUE #280: fxFlex="1 1 0" sets flex-basis to 100%
CSS doesn't require units if the value is 0. `fxFlex="1 1 0px"`(with units) works correctly and sets flex-basis to 0px. However, `fxFlex="1 1 0"`(without units) erroneously sets it to 100%. `flex: 1 1 0;` is valid CSS.

Some talk in #277 suggested it might be related to [flexbug#4](https://github.com/philipwalton/flexbugs/blob/master/README.md). If someone sets fxFlex="1 1 0" (without units), this API should set flex to 1 1 0px (with units) to satisfy flexbug#4.

### LINKED ISSUE #528: (api): fxFlex ignores basis value when using .lt-md
The following configuration `fxFlex="grow" fxFlex.lt-md="1 0 50%"` should set basis value to 50% when the screen size is less than 960px.

Only the shrink value changes.

*  [StackBlitz Demo](https://stackblitz.com/edit/angular-flex-layout-seed-ywjru2)

### LINKED ISSUE #534: Incorrect max-width: 0px; when using fxFlex="1 1 0px"
Bug.

Effect of generated CSS properties in case of `fxFlex="1 1 0px"` should match the one of `flex: 1 1 0px;`.

When you use `fxFlex="1 1 0px"`, `max-width: 0px;` is getting added to an element which breaks displaying of the element in some cases.

This is side-by-side demonstration of the issue with a simplified layout from my project:
[StackBlitz](https://stackblitz.com/edit/angular-flex-layout-fxfex-0px)

Tested in IE 11, Edge, FF 57, Chrome 63.0.3239.84.
Angular 5.1.1, Material 5.0.1, Win 10, TS 2.4.2.

CaerusKaru:
* Fix for when flex-basis is unitless and 0
* Fix for when no width/height is applied and flex-basis should be
  set

Fixes #277
Fixes #280
Fixes #528
Fixes #534

### DIFF
--- a/src/lib/api/flexbox/flex.spec.ts
+++ b/src/lib/api/flexbox/flex.spec.ts
@@ -142,7 +142,49 @@ describe('flex directive', () => {
       }
     });
 
-    it('should add correct styles for `fxFlex="0%"` usage', () => {
+    it('should add correct styles for flex-basis unitless 0 input', () => {
+      componentWithTemplate(`<div fxFlex="1 1 0"></div>`);
+
+      fixture.detectChanges();
+      expectNativeEl(fixture).toHaveStyle({
+        'flex': '1 1 0%',
+        'box-sizing': 'border-box',
+      }, styler);
+
+      expectNativeEl(fixture).not.toHaveStyle({
+        'max-width': '*'
+      }, styler);
+    });
+
+    it('should add correct styles for flex-basis 0px input', () => {
+      componentWithTemplate(`<div fxFlex="1 1 0px"></div>`);
+
+      fixture.detectChanges();
+      expectNativeEl(fixture).toHaveStyle({
+        'flex': '1 1 0%',
+        'box-sizing': 'border-box',
+      }, styler);
+
+      expectNativeEl(fixture).not.toHaveStyle({
+        'max-width': '*'
+      }, styler);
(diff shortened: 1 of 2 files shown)


CaerusKaru:
Note: this is the first of hopefully many PRs to address the backlog of flexbox issues.

CaerusKaru:
Closed because this was on the wrong branch

PR #631, opened 2018-02-26
refactor(lib): change package structure to allow for entrypoints

CaerusKaru:
* Add entrypoints for core, flex, and extended. All entrypoints have
  a module that can be imported separately
* Remove the BiDi clone from the CDK and add a direct/peer dependency


CaerusKaru:
@DevVersion Can you please review this?

ThomasBurleson:
Merged with SHA 32e6520

PR #633, opened 2018-02-27
fix(release): Fix release script by removing ',' of the last item in …

tinayuangao:
…the list


ThomasBurleson:
Thank you @tinayuangao

PR #639, opened 2018-03-04
build: reset build to use TypeScript 2.6.x

### LINKED ISSUE #638: AOT Builds fail with Angular 5.2.7
Delete `package-lock.json` and run `npm i && npm run lib:build:aot`
See failures in terminal.

> Works fine in Angular 5.2.6.

CaerusKaru:
* Lock version of TypeScript to 2.6.x since Angular doesn't support
  v2.7 yet
* Re-add MediaQueriesModule as a deprecated alias for CoreModule
* Fix AOT/Universal tasks to check for presence of lock file before
  deleting it

Fixes #638

### DIFF
--- a/src/apps/demo-app/package.json
+++ b/src/apps/demo-app/package.json
@@ -17,7 +17,7 @@
     "@angular/common": "file:../../../node_modules/@angular/common",
     "@angular/compiler": "file:../../../node_modules/@angular/compiler",
     "@angular/core": "file:../../../node_modules/@angular/core",
-    "@angular/flex-layout": "file:../../../dist/releases/flex-layout",
+    "@angular/flex-layout": "file:../../../dist/releases/flex-layout/angular-flex-layout.tgz",
     "@angular/forms": "file:../../../node_modules/@angular/forms",
     "@angular/http": "file:../../../node_modules/@angular/http",
     "@angular/material": "file:../../../node_modules/@angular/material",
--- a/src/lib/flex/flex-offset/flex-offset.ts
+++ b/src/lib/flex/flex-offset/flex-offset.ts
@@ -17,7 +17,13 @@ import {
   SkipSelf,
 } from '@angular/core';
 import {Directionality} from '@angular/cdk/bidi';
-import {BaseFxDirective, MediaChange, MediaMonitor, StyleUtils} from '@angular/flex-layout/core';
+import {
+  BaseFxDirective,
+  MediaChange,
+  MediaMonitor,
+  StyleDefinition,
+  StyleUtils,
+} from '@angular/flex-layout/core';
 import {Subscription} from 'rxjs/Subscription';
 
 import {LayoutDirective} from '../layout/layout';
@@ -156,7 +162,7 @@ export class FlexOffsetDirective extends BaseFxDirective implements OnInit, OnCh
     this._applyStyleToElement(this._buildCSS(value));
   }
 
-  protected _buildCSS(offset): { [x: string]: string } {
+  protected _buildCSS(offset): StyleDefinition {
     let isPercent = String(offset).indexOf('%') > -1;
     let isPx = String(offset).indexOf('px') > -1;
     if (!isPx && !isPercent && !isNaN(offset)) {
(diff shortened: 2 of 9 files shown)


src/lib/flex/flex-offset/flex-offset.ts
```diff
@@ -156,7 +162,7 @@ export class FlexOffsetDirective extends BaseFxDirective implements OnInit, OnCh
     this._applyStyleToElement(this._buildCSS(value));
   }
 
-  protected _buildCSS(offset): { [x: string]: string } {
+  protected _buildCSS(offset): StyleDefinition {
```
ThomasBurleson: Good improvement.

src/apps/demo-app/package.json
```diff
@@ -17,7 +17,7 @@
     "@angular/common": "file:../../../node_modules/@angular/common",
     "@angular/compiler": "file:../../../node_modules/@angular/compiler",
     "@angular/core": "file:../../../node_modules/@angular/core",
-    "@angular/flex-layout": "file:../../../dist/releases/flex-layout",
+    "@angular/flex-layout": "file:../../../dist/releases/flex-layout/angular-flex-layout.tgz",
```
ThomasBurleson: Why is a `.tgz` needed here ?

tools/gulp/tasks/aot.ts
```diff
@@ -1,20 +1,46 @@
...
+  'aot:bundle:rename',
+  'aot:clean',
+  'aot:deps',
+  'aot:add:tar')
+);
 
 task('aot:deps', [], execTask(
   'npm', ['install'], {cwd: demoAppSource}));
 
+task('aot:add:tar', [], execTask(
```
ThomasBurleson: Please add comment paragraph here to explain your intent.

tools/gulp/tasks/aot.ts
```diff
@@ -29,12 +55,13 @@ task('aot:clear:mods', [], execTask(
   }
 ));
 
-task('aot:clear:lock', [], execTask(
+task('aot:clear:lock', [], existsSync(join(demoAppSource, 'package-lock.json')) ?
+  execTask(
```
ThomasBurleson: Let's do this here also:

```ts
 () => existsSync(join(universalAppSource, 'package-lock.json')) ?
```

> Similar to the fix in Universal.

tools/gulp/tasks/universal.ts
```diff
@@ -74,12 +75,15 @@ task('prerender:clear:deps', [], execTask(
   }
 ));
 
-task('prerender:clear:lock', [], execTask(
-  'rm', ['./package-lock.json'], {
-    silent: true,
-    cwd: universalAppSource
-  }
-));
+task('prerender:clear:lock', [], () => existsSync(join(universalAppSource, 'package-lock.json')) ?
```
ThomasBurleson: Very nice!  
> I did not think of this approach.

src/apps/demo-app/package.json
CaerusKaru: To maintain parity with the Universal task, you'll see it has the same logic. There was some reason in the Universal app why direct directory access wasn't working (I think a mismatch in Angular versioning/linking issue), so installing it was the only way to get it to function correctly.

tools/gulp/tasks/aot.ts
CaerusKaru: As cited above, but will add to the file

tools/gulp/tasks/aot.ts
```diff
@@ -1,20 +1,53 @@
...
 
 task('aot:deps', [], execTask(
   'npm', ['install'], {cwd: demoAppSource}));
 
+
+/**
+ * The following tasks bundle Flex Layout into a tar file that can be installed
+ * directly instead of linked to via sym link. When linking, there are some issues
+ * that npm introduces, like Angular functionality impairment. This also has the
+ * benefit of better simulating the install process for Flex Layout in a CLI app
```
ThomasBurleson: Nice.

ThomasBurleson:
Merged with SHA 95a6e83

PR #640, opened 2018-03-04
refactor(lib): remove RendererAdapter from the library

CaerusKaru:
* Remove the RendererAdapter from NgStyle and NgClass extenders
  because Flex Layout no longer supports Angular v2


ThomasBurleson:
Merged with SHA 3354742

PR #643, opened 2018-03-07
Fix RegExp Issue in older versions of Safari

swftvsn:
Error is: TypeError: Cannot supply flags when constructing one RegExp from another.  

Discussion here: https://stackoverflow.com/questions/37919802/regex-created-with-constructor-not-working-with-safari

### DIFF
--- a/src/lib/core/browser-provider.ts
+++ b/src/lib/core/browser-provider.ts
@@ -17,7 +17,12 @@ export function removeStyles(_document: Document, platformId: Object) {
   return () => {
     if (isPlatformBrowser(platformId)) {
       const elements = Array.from(_document.querySelectorAll(`[class*=${CLASS_NAME}]`));
-      const classRegex = new RegExp(/\bflex-layout-.+?\b/, 'g');
+
+      // RegExp constructor should only be used if passing a variable to the constructor.
+      // When using static regular expression it is more performant to use reg exp literal.
+      // This is also needed to provide Safari 9 compatibility, please see
+      // https://stackoverflow.com/questions/37919802 for more discussion.
+      const classRegex = /\bflex-layout-.+?\b/g;
       elements.forEach(el => {
         el.classList.contains(`${CLASS_NAME}ssr`) && el.parentNode ?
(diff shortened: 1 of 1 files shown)


CaerusKaru:
I just tested that expression on both Safari 10 and 11. The issue was opened in 2016 and looks like it's been resolved in the latest two versions of Safari, which is all we support.

Thanks for the contribution, but it hardly seems necessary now.

swftvsn:
That's very unfortunate, as that is the single thing breaking safari 9 and older support :/ It's also a safe change, and the RegExp constructor does nothing but adds clutter and decreases performance. (As two regexps are constructed.)

swftvsn:
Hmm, I probably then have to fork & support a separate version for us.

CaerusKaru:
Alright, you've convinced me. I'll run this up to @ThomasBurleson but if it's minimal code and allows for Safari 9 support to be added back, I'm for it.

CaerusKaru:
@swftvsn please add a comment above the regex change explaining why it’s necessary

swftvsn:
Is the comment I added enough?

CaerusKaru:
@swftvsn Looks good, but @ThomasBurleson might suggest edits before it gets merged

CaerusKaru:
@swftvsn Looks like you've got some linting errors that need to be resolved. Other than that, tests look good.

swftvsn:
Fixed tslint errors.

CaerusKaru:
@swftvsn Pro tip: Just linking https://stackoverflow.com/questions/37919802 will resolve to the full URL

swftvsn:
Thanks!

ThomasBurleson:
@swftvsn - please squash your commits. Thank you for this fix.

swftvsn:
Haha, seems that I don't know how to do a squash. If I understand correctly the user interface of github allows this as part of the merge. However, I'm not allowed to merge, so I tried to pull my branch, and then do  

```
git rebase -i HEAD~4
git pull
git push
```
But I think it's worse now. So, can you provide any assistance what should I do with that branch next? And how do I provide CLA. The committer is me.

PR #652, opened 2018-03-10
build: add bazel rules

CaerusKaru:


CaerusKaru:
Blocked on #681

CaerusKaru:
This is failing because of a Material issue in the demo-app. Hoping they resolve this with their RxJS fix...

BUILD.bazel
```diff
@@ -0,0 +1,78 @@
+package(default_visibility = ["//visibility:public"])
+
+# TODO(jelbourn): figure out if these workarounds are still needed
```
alexeagle: don't add TODOs for Jeremy in this repo :)

src/lib/BUILD.bazel
```diff
@@ -0,0 +1,32 @@
...
+  deps = [
+    "//src/lib/core",
+    "//src/lib/extended",
+    "//src/lib/flex",
+    "//src/lib/server",
+  ],
+  tsconfig = ":tsconfig-build.json",
+)
+
+# Creates the @angular/flex-layout package published to npm.
```
alexeagle: Should probably add a comment explaining how to publish a release (or point to docs that say `bazel run //src/lib:npm_package.publish` )

CaerusKaru: Added

BUILD.bazel
CaerusKaru: Since this falls under Material, I'll keep it in, though it's not likely he'll ever see it 😉

CaerusKaru:
Extra commits OK with me @googlebot

src/lib/BUILD.bazel
ThomasBurleson: Added link to Bazel.md documentation also.

CaerusKaru:
This is blocked indefinitely as there are three main issues pending (mostly related to Angular Core):

1) Angular Core now pollutes the global namespace when testing in NodeJS mode. This conflicts with browser checking in the CDK Platform service and throws off the tests
2) Angular Core now requires a downgrade in `@types/jasmine` because we compile their testing utilities as a postinstall task for Bazel. This filters down to apps that consume our library because our testing utilities will now also be exposed
3) Bazel testing no longer works because Karma gets a script error due to a "Mismatched anonymous define() module" error. This will need further investigation, but it's likely due to how Angular is now bundled with Bazel instead of build.sh in their repo

cc @alexeagle for 2 and 3

alexeagle:
@CaerusKaru for 2 you should grab https://github.com/angular/angular/pull/22810/files#diff-93296510676816abb37409ccd9a174e0R20 - you should not need to `ngc` on the testing utilities

For 3, this error message needs improvement, it's saying that the `ts_web_test` was handed an anonymous AMD module which it doesn't currently support. I doubt this is because of Angular being published by Bazel - we should have all named modules both before and after the change. Can you figure out which module you pull in that is anonymous (first argument to `define` is an array rather than a string?) - this might be related to 2

CaerusKaru:
Closed in favor of #704

PR #658, opened 2018-03-12
fix(fxFlex): make sure not to set width/height when flex is default

CaerusKaru:
* fix an issue where when fxFlex value is blank, the max-width
  would be set to the default flex (1e-9px)
* move server utils into the Core module since all modules import
  the core module
* fix demo app serve method to rebuild the demo app correctly

### DIFF
--- a/src/apps/demo-app/package-lock.json
+++ b/src/apps/demo-app/package-lock.json
@@ -39,8 +39,8 @@
       }
     },
     "@angular/animations": {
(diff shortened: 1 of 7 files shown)


ThomasBurleson:
@CaerusKaru  - rebase needed?

CaerusKaru:
@ThomasBurleson hang on, will take care of this now (it overlapped a little with the lazy loading PR)

CaerusKaru:
@ThomasBurleson rebased, once CI passes this should be good to go!

PR #662, opened 2018-03-13
fix(fxFlex): fix wrapping in older versions of Safari

### LINKED ISSUE #660: "row wrap" issue in Safari 10.1, 9.1
#### Bug:
flex-basis shouldn't be set as 100%, should follow the value in fxFlex or max-width.

#### What is the Expected behavior?
 Wrap should work as earlier before "row wrap" incorporation instead of fxLayout="row" fxLayoutWrap?

#### What is Current behavior?
 It wraps child irrespective of child width if "row wrap" is set on the parent in below mentioned platform?

"@angular/core": "5.2.8"
"@angular/flex-layout": "5.0.0-beta.13"
"@angular/material": "5.2.4"
OS:Mac OSx,
Browser: Safari9.1, 10.1

For reference: 
flex-basis is being set 100% irrsepective of max-width on element which might be causing the issue.
eg.
fxFlex="55"
was tunring into 
earlier:
flex: 1 1 55%;
box-sizing: border-box;
display: flex;
max-width: 55%;

now
flex: 1 1 100%;
box-sizing: border-box;
display: flex;
max-width: 55%;

CaerusKaru:
* add flex basis back if either max or min width/height is set
* related: https://github.com/philipwalton/flexbugs#flexbug-11
* fix demo-app height issue with older Safari
* update tests to reflect new behavior
* update port for demo app to work with BrowserStack Local

Fixes #660 

BREAKING CHANGE:

* `fxFlex` now sets the flex basis to the `max-width/height` or `min-width/height` if they are set

### DIFF
--- a/src/apps/demo-app/src/app/github-issues/issue-266/issue-266.component.html
+++ b/src/apps/demo-app/src/app/github-issues/issue-266/issue-266.component.html
@@ -17,7 +17,7 @@
         <div class="handle handle-row" ngxSplitHandle>
           <i class="material-icons">&#xE25D;</i>
         </div>
-        <div fxFlex.xs="70%" fxFlex.gt-md="50%" fxFlex.lg="60%" ngxSplitArea>
+        <div fxFlex.xs="70%" fxFlex.gt-md="50%" fxFlex.lg="60%" ngxSplitArea style="height:500px;">
           <div fxLayout="column" fxFlexFill ngxSplit="column">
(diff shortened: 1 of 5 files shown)


CaerusKaru:
Blocked on #658

ThomasBurleson:
Rebase after 658 merge?

CaerusKaru:
@ThomasBurleson rebased

PR #667, opened 2018-03-15
feat(fxLayoutGap): add gutter functionality to layout-gap

### LINKED ISSUE #134: better way of handling gap/gutter
Currently when you want to have a grid of items with gutter between them, this happens https://plnkr.co/edit/eEWOqBJA4rq0PFKl1XyT?p=preview (unless there's some different hidden way of doing gaps other than the obvious `fxLayoutGap`.. I swear I've seen some `calc` example somewhere, but can't find it again).. 

image

but ideally you'd want 4 items(25% each) in a row like this
image

There's a simple method for this which has been used for grid systems all over the place(and works fine with flex as well) and that is using negative margin on the container and padding on items(see https://plnkr.co/edit/ZhNvWutftQ1qRMO6cHzZ?p=preview) 

the only real downside is that you have to use `fxFlex` elements only as a "container" for your content(ie. you can't add background or horizontal paddings to them)... but it's much more performant than say calc and much more usable than the way `fxLayoutGap` currently works. Could it perhaps be added as a new attribute, say `fxLayoutGutter`?

### LINKED ISSUE #235: Shouldn't fxFlex value be independent of fxLayoutGap (with fxLayout="row")
The problem I am trying to highlight is the relation between the `fxLayoutGap` and the `fxFlex` percentage used with `fxLayout="row"`.
To me, the `fxFlex` value should a percentage of the available space minus the `fxLayoutGap` applied inbetween the elements.

So in the following screenshot for example, I would expect to have 2 lines of 4 squares (because 25% is a quarter of 100%):
image

https://plnkr.co/edit/W6N0q7mSjEjBKGyws3R4?p=preview

### LINKED ISSUE #288: Space between fxFlex not work with fxLayoutWrap and fxLayoutGap?
I'm using layout-flex to format the fields of a form and in this format I have the following problem:

sem titulo

Code:

Should not the spacing applied by fxLayoutGap be between the elements? So in the example above what should happen would not be a reduction of the first two fxFlex = "50" to contain the spacing and still mater in only one line?

CaerusKaru:
* Apply negative margin and positive padding as browser-agnostic
  hack for grid gutter
* The new feature is activated by appending `grid` to any
  `fxLayoutGap` input, e.g.
  `fxLayoutGap="10px grid"`

Fixes #134 
Fixes #235 
Fixes #288

### DIFF
--- a/src/apps/demo-app/src/app/layout/docs-layout/docs-layout.component.ts
+++ b/src/apps/demo-app/src/app/layout/docs-layout/docs-layout.component.ts
@@ -5,6 +5,7 @@ import {Component} from '@angular/core';
   template: `
     <demo-layout-alignment class="small-demo"></demo-layout-alignment>
     <demo-layout-fill class="small-demo"></demo-layout-fill>
+    <demo-layout-gap class="small-demo"></demo-layout-gap>
     <demo-flex-row-fill class="small-demo"></demo-flex-row-fill>
     <demo-flex-row-fill-wrap class="small-demo"></demo-flex-row-fill-wrap>
     <demo-flex-attribute-values class="small-demo"></demo-flex-attribute-values>
--- a/src/apps/demo-app/src/app/layout/layout-gap/layout-gap.component.spec.ts
+++ b/src/apps/demo-app/src/app/layout/layout-gap/layout-gap.component.spec.ts
@@ -1,0 +1,25 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { LayoutGapComponent } from './layout-gap.component';
+
+describe('LayoutGapComponent', () => {
+  let component: LayoutGapComponent;
+  let fixture: ComponentFixture<LayoutGapComponent>;
+
+  beforeEach(async(() => {
+    TestBed.configureTestingModule({
+      declarations: [ LayoutGapComponent ]
+    })
+    .compileComponents();
+  }));
+
+  beforeEach(() => {
+    fixture = TestBed.createComponent(LayoutGapComponent);
+    component = fixture.componentInstance;
+    fixture.detectChanges();
+  });
+
+  it('should create', () => {
+    expect(component).toBeTruthy();
+  });
+});
--- a/src/apps/demo-app/src/app/layout/layout-gap/layout-gap.component.ts
+++ b/src/apps/demo-app/src/app/layout/layout-gap/layout-gap.component.ts
@@ -1,0 +1,70 @@
+import {Component} from '@angular/core';
+
+const DIRECTIONS = ['row', 'row-reverse', 'column', 'column-reverse'];
+
+@Component({
+  selector: 'demo-layout-gap',
+  template: `
+    <mat-card class="card-demo">
+      <mat-card-title><a href="" target="_blank">Layout Gap</a></mat-card-title>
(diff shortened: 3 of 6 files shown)


ThomasBurleson:
Why is this a better solution?

CaerusKaru:
@ThomasBurleson I've added a demo if you want to run and see for yourself. The problem is that the current solution only provides a margin to add a "gap" between elements, and requires hacks to get the sizing right (e.g. `fxFlex="25"` four times ends up spilling over).

The proposed solution adds the desired behavior by not only properly sizing the elements, but also providing a correct gutter between elements in multiple sections.

The demo very clearly shows this, I highly encourage you take a look for yourself to see.

CaerusKaru:
So this is a little bit of a hacky solution. Maybe we should add this as another directive to augment LayoutGap? We could call it LayoutGapGrid?

CaerusKaru:
The angle I've settled on is that instead of introducing a breaking change and basically rewriting `fxLayoutGap`, and instead of adding a new directive, this could be a new option on `fxLayoutGap`.

The new proposed syntax would be:
`fxLayoutGap = <gap> [grid]`, where grid indicates that the new stylings should be used in place of the default styles.

I'll change this PR to reflect this and we can move from there.

intellix:
Just wanna say I'm in favour of helpers for negative margins etc. I'm currently doing this throughout all my projects over and over again to achieve this (along with fillers to maintain consistent box sizes when wrapping:

```html
<div class="flex-grid" fxLayout="row wrap">
  <xc-item [item]="item" fxFlex="1 0 16rem" *ngFor="let item of items$ | async; trackBy: getObjectIdentity"></xc-item>
  <div class="flex-filler" fxFlex="1 0 16rem" *ngFor="let i of [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]"></div>
</div>
```

```scss
.flex-grid {
  margin: 0 -1rem -1rem 0;
}

xc-item {
  margin: 0 1rem 1rem 0;
}

.flex-filler {
  margin: 0 1rem 0 0;
}
````

**Comment by CaerusKaru** [comment]
@intellix So that's a vote in favor, right? 😉

**Comment by simeyla** [comment]
@CaerusKaru Where's the documentation for this? Thanks
https://github.com/angular/flex-layout/wiki/fxLayoutGap-API

Also any documentation needs to mention the possibility of requiring an additional wrapper with `overflow: hidden` if the flex container is interfering with its parent (eg. for full page width where you don't want scrollbars).

**Comment by AnthonyNahas** [comment]
This feature is really awesomeeeeee!!! 

adding the grip word next the the gap value fixed my layout in the [@angular-material-extensions/calendar](https://github.com/angular-material-extensions/calendar) library like the following 👍: 

without the grid
```
<div fxLayout="row wrap"
     fxLayout.lt-sm="column"
     fxLayoutGap="32px"
     fxLayoutAlign="center">

.....
```
bildschirmfoto 2018-11-11 um 06 59 16


with grid
```
<div fxLayout="row wrap"
     fxLayout.lt-sm="column"
     fxLayoutGap="32px grid"
     fxLayoutAlign="center">
```
bildschirmfoto 2018-11-11 um 06 56 54

Estevete:
It does not work in Microsoft Edge. How can I make it work?

PR #669, opened 2018-03-16
fix(fxFlex): fix non-wrapping behavior and default fxFlex value

### LINKED ISSUE #314: Column Layout Breaks in IE11
I opened a similar issue earlier this year about this. It was closed as it was a question, but it wasn't: I really think it's a bug, or, at least, an unforseen behavior in IE11. A column layout is working very diferent in IE11 and in Chrome. In fact it seems it is broken in IE11, as you can see in the two images bellow.

In IE11 the item in a `fxLayout="column"` container aparently collapses if you do both of this simultaneously:

- don't specify a height for the container;
- use `fxFlex` attribute in the contained items without a fixed height: it's not possible to leave it empy - just `fxFlex` -  or set a percentage value - for example `fxFlex="40%"` or `fxFlex="40"`. 

I think, in this case, in IE11, the behavior should be the same as the` fxFlex="nogrow"` option, as it is in other browsers.

plunker: https://plnkr.co/edit/zGMk1PaG5lCGsx2YAp79?p=preview

## **Chrome:**
image

## **IE11:**
image

### LINKED ISSUE #338: property:flex parsing incorrect value `1e-09px;`
property:flex parsing incorrect value `1e-09px;`  
flex

Googling 1e-09 seems to be some kind of scientific notation, and can be see as a % , / in some code out there.

`1 n  1E-09
1 µ  1E-06
1 m  1E-03
1    1
1 k  1E+03
1 M  1E+06
1 G  1E+09`

### LINKED ISSUE #481: Responsive Flex Layout Chrome Bug
Bug

When moving from row to column layout, controls should stack appropriately

In Chrome, when transitioning from row to column layout with an element that has a nested column layout, all of the elements merge onto the same line.

##### Examples  

*Chrome at gt-md*  
chrome-gt-md  

*Chrome at lt-lg*  
chrome-lt-lg  

*Firefox at lt-lg*  
firefox-lt-lg  

[Repro](https://stackblitz.com/edit/angular-flex-layout-seed-dwwepm)  
*All of the relevant markup is in test.component.html*

Tested with the following:
Angular 4.4.4
Material 2 beta 12
flex-layout 2 beta 10

While the following causes the mess in Chrome, but works in Firefox and Edge:  

``` html
<div fxFlex fxLayout="column" fxLayoutAlign="start stretch">
```

The following causes the intended behavior to work in all browsers:  

``` html
<div fxLayout="column" fxLayoutAlign="start stretch" [ngStyle.gt-md]="{'flex': '1 1 auto'}">
```

### LINKED ISSUE #544: sticky footer issue with flex-basis
Checking my app on IE10-11 uncovered I was see an issue with my sticky footer. It was not sticky, but just say under the content block positioned above:
```
+site(fxLayout="column" fxFlex)
  |- header
  |- content(fxFlex)
  |- footer
==white-space==
```
So I went back to the article written by Philip Walton a couple of years ago, https://philipwalton.com/articles/normalizing-cross-browser-flexbox-bugs/, and implemented his fix.

However, I was a bit puzzled that this fix is still required nowadays.
So I was wondering if we still need to do this:

```
+site(fxLayout="column" fxFlex) --> add height: 100%
  |- header  --> add flex-shrink: 0
  |- content(fxFlex)  --> add flex: 1 0 auto
  |- footer -->add flex-shrink: 0
```
Only if I add the `flex:1 0 auto` to the content block, will it work fine in Chrome. I am actually overriding fxFlex at this moment.

Header at top, content stretched and footer sticky at the bottom.

In IE10-11 footer is not sticky at bottom. When adding height 100% to `site`, then Chrome will squeeze the content block so that everything will fit. Fix on, break the other and vice versa.

### LINKED ISSUE #666: bug(fxFlex): flex basis should only be set to width value when wrapping
Currently the flex-basis is set to the value of the width when present as a fix for wrapping in older versions of Safari. However, this throws off ordinary flex values when not wrapping (and is not needed when not wrapping).

Need a partial revert of the recent PR to fix this, with a patch that detects wrapping

CaerusKaru:
* Set the non-wrapping behavior for fxFlex to be 100% instead of
  the width/height value
* Set the default value for fxFlex to auto for column and 0% for
  row instead of 1e-9px

Fixes #666 
Fixes #481 
Fixes #544 
Fixes #338 
Fixes #314

### DIFF
--- a/src/lib/flex/flex-offset/flex-offset.spec.ts
+++ b/src/lib/flex/flex-offset/flex-offset.spec.ts
@@ -8,7 +8,6 @@
 import {Component, PLATFORM_ID} from '@angular/core';
 import {CommonModule, isPlatformServer} from '@angular/common';
 import {ComponentFixture, inject, TestBed} from '@angular/core/testing';
-import {Platform, PlatformModule} from '@angular/cdk/platform';
 import {DIR_DOCUMENT} from '@angular/cdk/bidi';
 import {SERVER_TOKEN, StyleUtils} from '@angular/flex-layout/core';
 
@@ -25,15 +24,13 @@ describe('flex-offset directive', () => {
   let fixture: ComponentFixture<any>;
   let fakeDocument: {body: {dir?: string}, documentElement: {dir?: string}};
   let styler: StyleUtils;
-  let platform: Platform;
   let platformId: Object;
   let componentWithTemplate = (template: string) => {
     fixture = makeCreateTestComponent(() => TestFlexComponent)(template);
 
-    inject([StyleUtils, Platform, PLATFORM_ID],
-      (_styler: StyleUtils, _platform: Platform, _platformId: Object) => {
+    inject([StyleUtils, PLATFORM_ID],
+      (_styler: StyleUtils, _platformId: Object) => {
       styler = _styler;
-      platform = _platform;
       platformId = _platformId;
     })();
   };
@@ -44,7 +41,7 @@ describe('flex-offset directive', () => {
 
     // Configure testbed to prepare services
     TestBed.configureTestingModule({
-      imports: [CommonModule, FlexLayoutModule, PlatformModule],
+      imports: [CommonModule, FlexLayoutModule],
       declarations: [TestFlexComponent],
       providers: [
         {provide: DIR_DOCUMENT, useValue: fakeDocument},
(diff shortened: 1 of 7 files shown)


CaerusKaru:
Note: confirmed issue with IE11 by removing the hack. Will consult Flexbugs and try to come up with an alternative that meets all requirements.

CaerusKaru:
Fixed the default value and officially unblocked

ThomasBurleson:
rebase this one also plz @CaerusKaru

CaerusKaru:
@ThomasBurleson done

PR #670, opened 2018-03-16
feat(tokens): add configuration for breakpoints and flex styles

### LINKED ISSUE #263: fxFlex directive adding styles to both parent and host element
When `fxFlex` is added to a child element whose parent does not contain `fxLayout`, the library is automatically adding inline styles to the parent element (as well as the host) that essentially convert the parent to a flex row.

While I understand the documentation states:

> The fxFlex directive should be used on elements within a fxLayout container and identifies the resizing of that element within the flexbox container flow.

There are valid use-cases that require the use of fxFlex to be used while not inside an `fxLayout` container, and the expected result would be a simply application of `flex: 1 1 auto` or whatever the desired style is on the **host** element, while leaving the **parent** element alone.

One such example would be a routing component that needs to be a flex column. Currently, I have styles in my component stylesheet that set it to be a flex column on the `:host`. Then the child elements inside the component template use `fxFlex` to stretch to fill that container. Unfortunately `flex-layout` is currently converting those routed components to be a flex **row** with inline styles that clobber the component stylesheet.

The desired behaviour would therefore be to leave the parent component of the host untouched.

Originally posted here: https://github.com/angular/angular/issues/15726

### LINKED ISSUE #496: `flex-direction: row; box-sizing: border-box; display: flex;` added automatically
I am seeing `flex-direction: row; box-sizing: border-box; display: flex;`added on my component that is rendered in the router-outlet.

My app.component.html looks like
```
<pro-header *ngIf="!headless"></pro-header>
<router-outlet></router-outlet>
<pro-footer *ngIf="!headless"></pro-footer>
```
The router-outlet will render a component which I give a class `fxFlex`
via 
```
@HostBinding( 'class.fxFlex') true;
```
so it is flexed.
However, when the components root DIV uses fxLayout column WITH an fxFlex, like
```
<div fxLayout="column" fxLayoutAlign="center center" fxFlex>
```
it automatically adds flex-direction row to its parent.
The rendered html will be like
```
<pro-header *ngIf="!headless"></pro-header>
<router-outlet></router-outlet>
<pro-securing style="flex-direction: row; box-sizing: border-box; display: flex;">
....
</pro-securing>
<pro-footer *ngIf="!headless"></pro-footer>
```
This row layout on the `pro-securing` component creates the side effect that when the footer becomes large, it pushes the pro-securing into the header. It doesn't allow a scrollbar.

I find this a bit strange behaviour.
I am looking for a header, content (can expand) and sticky footer. Scroll bar should automatically engage when content is larger then header+content+footer.

Any thoughts?

### LINKED ISSUE #518: When using custom breakpoints, can't use defaults ones anymore
Bug

After having added custom breakpoints, we should be able to use the default breakpoints. In addition, if a DOM element has a style attribute defined, it should be used if no style is enforced by the flex-layout package.

**fxLayout**, **ngStyle** and **ngClass** work with the custom breakpoints, but not with the default breakpoints (_xs_, _sm_, _md_, _lg_, _xl_). In addition, if you define the attributes _style_, and _ngStyle.custom_breakpoint_, once the rules for _custom_breakpoint_ have been activated, _style_ is not applied anymore (as it should be when _custom_breakpoint_ is no longer active).

A Plunker is available [here](https://plnkr.co/edit/2IWMsmL3KjKm6owFA3Dj?p=preview).

Have the directives work as expected and presented in the doc.
Avoid defining the `ObservableMedia` in each component, and writing all style-related rules with conditions like `media.isActive('custom_breakpoint')`.

Tested with Angular5.0.2 and 5.0.3, on Windows 10, with Typescript 2.3.4 and 2.6.2, using Firefox and Chrome.
Tested with the latest beta of flex-layout.

### LINKED ISSUE #575: Configurable vendor prefixes
Feature request

Would like to disable vendor prefixes during dev. I don't need them at all right now so making it configurable would be very nice.

Always enabled.

It's painful while in chrome dev tools and i want to toggle a flex property. I have to toggle all prefixes because they act as fallbacks for the regular css property i am trying to toggle. It's actually faster to change in code and wait for rebuild sometimes. Where clicking one checkbox in dev tools would be greatly appreciated.

Thanks for your help, and this crazy cool lib. Had no idea it would be considerably more powerful than ng-material1's layout. Pretty impressive.

CaerusKaru:
* add token to add individual breakpoints that can be merged with
  the defaults seamlessly
* add token that optionally adds flex stylings to parents without
  inline flex styles (set to false by default)
* add token to optionally disable adding vendor prefixes to inline
  styles

BREAKING CHANGE:

* `fxFlex` no longer adds `display: flex; flex-direction: row` by
  default

Fixes #496 
Fixes #518 
Fixes #575 
Fixes #263


CaerusKaru:
Note: this probably needs two more tests before it's ready (one for the new `BREAKPOINT` token, and one for the new `DISABLE_VENDOR_PREFIXES` token)

PR #680, opened 2018-03-20
fix(breakpoints): add global provider for BreakPointRegistry

CaerusKaru:
* Fix issue with breakpoints not activating because the new
  breakpoints token is emptied when BreakPointRegistry is
  injected again

### DIFF
--- a/src/lib/core/breakpoints/break-point-registry-provider.ts
+++ b/src/lib/core/breakpoints/break-point-registry-provider.ts
@@ -1,0 +1,33 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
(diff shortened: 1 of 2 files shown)


src/apps/demo-app/src/app/layout/layout-gap/layout-gap.component.ts
```diff
@@ -11,7 +11,7 @@ const DIRECTIONS = ['row', 'row-reverse', 'column', 'column-reverse'];
       </mat-card-subtitle>
       <mat-card-content class="large">
         <div fxFlexFill>
-          <div fxFlex
+          <div fxFlexFill
```
ThomasBurleson: Let's not include this change in this PR. This mixes fixes.

CaerusKaru: Fixed

PR #681, opened 2018-03-20
build: upgrade to Angular v6 and TypeScript v2.7

CaerusKaru:
* Upgrade to Angular v6 rc
* Upgrade to TypeScript v2.7
* Upgrade to RxJS v6 beta
* Upgrade to Material/CDK v6 beta
* Switch to tree-shakeable providers
* Remove beta 15 deletion targets
* Remove static method from main module


CaerusKaru:
Note: for some reason this increases the bundle size by 3kB. Seems odd, but might be due to how v6 bundles an app now. Regardless, this is only about a 1% increase (which may/may not be attributable to Flex Layout), so ignoring for now.

CaerusKaru:
Note: SSR tests currently failing due to non-specific error with Material code. Should be patched in the next Material release (likely when they upgrade to RxJS 6), and doesn't affect Flex Layout SSR. This also includes rxjs-compat until Material merges their RxJS PR

CaerusKaru:
Also note: the upgrade to RxJS v6 has led to a 40% increase in bundle size. I don't think there's much we can do about this since it's the way the boat is floating, but it's important to note regardless.

CaerusKaru:
Final note for the night: while the hello-world bundle goes up, the demo-app bundle almost uniformly decreases in size...

CaerusKaru:
Fixed SSR tests by changing dependencies, this PR is all set 👍

mhosman:
When you are gonna take this pull request? It would be great to test new Angular 6 rc0 with this update. Thanks!

PR #704, opened 2018-04-11
build: add bazel rules

CaerusKaru:
Continuation of #652


CaerusKaru:
cc @alexeagle for final review, especially the new CircleCI tasks that replace some of the ones that were on Travis (note: this is a redo of the last PR and is mostly the same, except smaller)

PR #721, opened 2018-04-24
test(grid-gap): fix chrome 66 naming issue

Teamop:
In the Chrome, [grid-gap becomes gap since v66](https://www.chromestatus.com/feature/4986266210795520), so need to fix the gap's unit test issue to pass the unit tests


CaerusKaru:
This looks great in principle, but since we only test against Chrome 66, can you change your PR to just check for Chrome (without version number)?

Teamop:
Wow, so quick response, thanks @CaerusKaru

PR #727, opened 2018-05-03
chore: bump deps to Angular and CDK v6 final

CaerusKaru:


fxck:
when can we expect v6 final of flex-layout? or are there not plans to have this lib aligned the same way material / cdk will be?

PR #729, opened 2018-05-03
feat(lib): add config options for flex basis and other tokens

CaerusKaru:
* Revert default flex-basis value in column mode back to 1e-9px
* Add static method on main module to allow for configuration of
  tokens, including the ability to set the flex-basis value in
  column mode to auto instead of the above hack
* Add flex stylings to parent to prevent excessive breaking change,
  also configurable by the static method
* Add extra check to prevent max-width from being added when
  `fxFlex` is empty
* Change development port to 5000
* Change karma reporter back to dots to get back detailed reporting
  in case of error

BREAKING CHANGE:

* Four configuration tokens have been removed:

  * `ADD_FLEX_STYLES`
  * `ADD_ORIENTATION_BREAKPOINTS`
  * `DISABLE_DEFAULT_BREAKPOINTS`
  * `DISABLE_VENDOR_PREFIXES`

  These tokens have been consolidated into a new configuration token:

  * `LAYOUT_CONFIG`

* The default column flex-basis has been reverted to `1e-9px`. To
  have that value be `auto` instead, set the config in your top-level
  module as follows:

```
FlexLayoutModule.withConfig({useColumnBasisZero: false})
```


bastienlemaitre:
Any new ?

ChrisProlls:
One question about the `.withConfig()` method : where do we put this code ? Into our top-level module ? Because I encounter an error : `The property 'provide' is missing in type 'ModuleWithProviders'.`

```TS
@NgModule({
  declarations: [
    AppComponent
  ],
  imports: [
    BrowserModule
  ],
  providers: [
    FlexLayoutModule.withConfig(FlexConfig, CustomBreakpoints)
  ],
  bootstrap: [AppComponent]
})
export class AppModule { }
```

PR #736, opened 2018-05-08
fix(style): apply correct flex with ngStyle and layout change

### LINKED ISSUE #700: fxFlex is not work well with NgStyle.lt-md and layout change
Set fxFlex="5 1 auto" with [ngStyle.lt-md]="{'width.px': 90}", but in the final html, got style: `flex: 1 1 0%`
the flex should be the same in the html

got style: `flex: 1 1 0%`

https://stackblitz.com/edit/angular-material-flex-layout-seed-76pec8, check the checkbox, then inspect the displayed span item next to it(items selected) 
*  [StackBlitz starter](https://goo.gl/y38GUZ) using latest `npm` nightly release: 

the unexpected style change

Angular 5.2.9
Angular material 5.2.4
Typescript 2.6.2

After some research, checking the checkbox triggers the layout change, so [_onLayoutChange() is invoked in flex](https://github.com/angular/flex-layout/blob/master/src/lib/flex/flex/flex.ts#L103) , then set `ngStyle` to `flex: 1 1 0%`, then for `NgStyle.lt-md`, the code `styles = extendObject({}, this._base.inputMap['ngStyle'], styles);` (https://github.com/angular/flex-layout/blob/master/src/lib/extended/style/style.ts#L179) merges this style to override the original style

Teamop:
Fix #700
For this issue, I found that the flex is initialized to '1 1 0%', when the layout changes, the layout event triggers before ngStyle, so invokes _updateStyle to apply this initialization to the ngStyleBase. Then for the following ngStyle.lt-md event, just merge this base style.

So the change is: when the input changes in the StyleDirective, check the base style, if it's not same with the current style, replace it with the current style and for others which have been merged with ngStyleBase, delete the corresponding style and merge again.

In the following test cases, the first case is just like the scenario in the #700

### DIFF
--- a/src/lib/extended/style/style.ts
+++ b/src/lib/extended/style/style.ts
@@ -35,7 +35,8 @@ import {
   NgStyleRawList,
   NgStyleType,
   NgStyleSanitizer,
-  ngStyleUtils as _
+  ngStyleUtils as _,
+  NgStyleMap
 } from './style-transforms';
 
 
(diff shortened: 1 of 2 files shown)


src/lib/extended/style/style.ts
```diff
@@ -109,6 +110,22 @@ export class StyleDirective extends BaseDirective
     if (this._base.activeKey in changes) {
       this._ngStyleInstance.ngStyle = this._base.mqActivation.activatedInput || '';
     }
+
+    // Fix for issue 700
```
praveeno: a hyperlink of issue is more useful perhaps,

praveeno: In my perspective, it leads confusion

CaerusKaru:
@Teamop this was a really great effort to track down this issue, but ultimately a lot of this is unnecessary. The real root of the problem (which, granted, was really tricky to find), was that the `ngIf` on the element was throwing off the directive lifecycles. There was some logic in `FlexDirective` that was too eager in initializing, thus using the default flex value instead of the real input.

All of this is encapsulated in #771, and I have taken the liberty of using your unit tests to confirm the behavior. Thank you again for your great work on this issue.

Teamop:
thanks @CaerusKaru . When I tried to find this issue, I thought about the lifecycle issue once, but I missed it finally 😭 .  Your solution is much better.

PR #779, opened 2018-06-13
fix(module): make withConfig AOT compatible

### LINKED ISSUE #778: AOT build throwing the error: Function calls are not supported in decorators but ‘FlexLayoutModule’ was called, when using withConfig options
# Bug Report

App should build in AOT when FlexLayoutModule.withConfig({addFlexToParent: false}) is added to the module.

App fails to build in AOT with error:  Function calls are not supported in decorators but ‘FlexLayoutModule’ was called, when using withConfig options

include in Module: FlexLayoutModule.withConfig({addFlexToParent: false})

Angular 6.0.4, Material 6.2.1

We have another module:     LocalStorageModule.withConfig({ prefix: '', storageType: 'localStorage' }) that builds fine.

CaerusKaru:
Fixes #778

### DIFF
--- a/src/apps/demo-app/src/app/app.module.ts
+++ b/src/apps/demo-app/src/app/app.module.ts
@@ -23,7 +23,7 @@ const PRINT_BREAKPOINTS = [{
     BrowserAnimationsModule,
     RoutingModule,
     DemoMaterialModule,
-    FlexLayoutModule,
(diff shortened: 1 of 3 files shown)


ChrisProlls:
So ... any news ? Will this PR be merged for the next release ?

Issue #822, opened 2018-08-25
SSR crashes due to undefined method


arturovt:
# Bug Report

#### What is the expected behavior?
`forEach` should be defined

#### What is the current behavior?
Angular Universal crashes with an error
`ERROR TypeError: Cannot read property 'forEach' of undefined at ResponsiveActivation._configureChangeObservers`

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
Angular 6.1.0
Flex Layout 6.0.0-beta.17

#### Is there anything else we should know?
I added `console.log(typeof this._registryMap)` in `node_modules/@angular/flex-layout/bundles/flex-layout-core.umd.js`, it printed `undefined` and `object` (on the client), because `ResponsiveActivation.prototype._buildRegistryMap` returns an array only in the case if `mediaMonitor` is not `undefined`

CaerusKaru:
Can you provide a minimal reproduction repository? We have SSR testing as part of our CI and it hasn't created this error.

arturovt:
@CaerusKaru Yeah, I'll try

PR #823, opened 2018-08-25
fix(core): add a check to an existing method

arturovt:
Fixes [#822](https://github.com/angular/flex-layout/issues/822)


CaerusKaru:
As mentioned in #822, we need a reproduction to be sure this is necessary.

Issue #846, opened 2018-09-26
withConfig requires breakpoints with AOT


gatimus:
# Bug Report

#### What is the expected behavior?
`FlexLayoutModule.withConfig({useColumnBasisZero: false})` as suggested [here](https://github.com/angular/flex-layout/blob/master/CHANGELOG.md#600-beta16-2018-06-07) to work with AOT compile

#### What is the current behavior?
produces
> Cannot read property 'alias' of undefined

#### What are the steps to reproduce?
https://github.com/gatimus/flex-layout-demo
`npm i`
`ng serve`

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
    Angular CLI: 6.2.3
    Node: 8.12.0
    OS: win32 x64
    Angular: 6.1.8
    ... animations, common, compiler, compiler-cli, core, forms
    ... http, language-service, platform-browser
    ... platform-browser-dynamic, router

    Package                           Version
    -----------------------------------------------------------
    @angular-devkit/architect         0.8.3
    @angular-devkit/build-angular     0.8.3
    @angular-devkit/build-optimizer   0.8.3
    @angular-devkit/build-webpack     0.8.3
    @angular-devkit/core              0.8.3
    @angular-devkit/schematics        0.8.3
    @angular/cdk                      6.4.7
    @angular/cli                      6.2.3
    @angular/flex-layout              6.0.0-beta.17
    @ngtools/webpack                  6.2.3
    @schematics/angular               0.8.3
    @schematics/update                0.8.3
    rxjs                              6.2.2
    typescript                        2.9.2
    webpack                           4.20.2

#### Is there anything else we should know?
`FlexLayoutModule.withConfig({ useColumnBasisZero: false }, DEFAULT_BREAKPOINTS)` works

CaerusKaru:
Please provide a minimal reproduction. This does not come up in our testing, which uses AOT.

CaerusKaru:
I had no issue running both `ng build --prod` and `ng serve`

CaerusKaru:
I see what you're seeing now. Let me dive into this and see if there's an easy fix.

PR #853, opened 2018-10-03
fix(module): do not require breakpoints in withConfig

livthomas:
I have fixed issue #846 that causes runtime errors when using `withConfig` without providing `breakpoints` parameter.

There is probably a better way to fix this by locating the root cause but this solution should work for now since it is a workaround I use in my application.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ node_modules
 # IDEs
 /.idea
 /.vscode
+/*.iml
 
 # misc
 .DS_Store
--- a/src/lib/module.ts
+++ b/src/lib/module.ts
@@ -41,28 +41,18 @@ export class FlexLayoutModule {
    * which sets the corresponding tokens accordingly
    */
   static withConfig(configOptions: LayoutConfigOptions,
-                    breakpoints?: BreakPoint|BreakPoint[]): ModuleWithProviders {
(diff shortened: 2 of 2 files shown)


CaerusKaru:
The real problem is that `breakpoint-token.ts` is currently:
```
export const BREAKPOINT = new InjectionToken<BreakPoint|BreakPoint[]|null>(
  'Flex Layout token, collect all breakpoints into one provider', {
    providedIn: 'root',
    factory: () => null
  });
```

when it should be 

```
export const BREAKPOINT = new InjectionToken<BreakPoint|BreakPoint[]|null>(
  'Flex Layout token, collect all breakpoints into one provider', {
    providedIn: 'root',
    factory: () => []
  });
```

I just have to patch it, but if you'd like to make the change and confirm it works, that's fine too.

CaerusKaru:
Ok, the real root cause for this:

We use `Array.isArray` on the `breakpoints` argument, but this function call doesn't work with AOT, and so we end up in the check as if it were an array (when it's really undefined), leading to errors downstream.

You've got part 1 of the fix (great work!), part 2 is to remove the `Array.isArray` check because it is no longer needed.

Again, excellent job!

livthomas:
@CaerusKaru Thanks for the explanation. I have removed the `Array.isArray` check.

PR #877, opened 2018-11-02
fix(fxLayoutAlign): Do not apply cross-axis strech styles when not ne…

### LINKED ISSUE #876: fxLayoutAlign sets wrong max-width when used with fxLayout
# Bug Report

```
<div fxLayout="column" fxLayoutAlign="center center"></div>
```
Should generate css:
```
{
    flex-direction: column;
    box-sizing: border-box;
    display: flex;
    place-content: center;
    align-items: center;
}
```

It adds an extra css property: `max-width: 100%`
```
{
    flex-direction: column;
    box-sizing: border-box;
    display: flex;
    max-width: 100%;
    place-content: center;
    align-items: center;
}
```

https://stackblitz.com/edit/angular-7ktrke?file=src%2Fapp%2Fapp.component.ts

[Documentation](https://github.com/angular/flex-layout/wiki/fxLayoutAlign-API) states that `max-width` should be set only if the cross axis is set to `stretch`.

Angular 7.0.2
flex-layout 7.0.0-beta.19

The problem only apears when both `fxLayout` and `fxLayoutAlign` are used together; using just `fxLayoutAlign` generates correct css.

The problem is that [`LayoutAlignDirective._allowStretching`](https://github.com/angular/flex-layout/blob/master/src/lib/flex/layout-align/layout-align.ts#L205) is called twice, the first time with `start stretch` - the default setting, and only after that with `center center`. 
`LayoutAlignDirective` does not reset the `max-width`/`max-height` when changing away from the `stretch` cross axis.

The first call is triggered by `ngOnChanges` on `LayoutDirective`, which triggeres `LayoutAlignDirective._onLayoutChange` on a LayoutAlign directive that is not yet initialized with the input values, so it uses the defaults `start stretch`.

The next call is triggered by `ngOnChanges` on `LayoutAlignDirective`, and this time the align settings are initialized correctly with `center center`, but `max-width` is never reset

danmana:
…eded. Closes #876

### DIFF
--- a/src/lib/flex/layout-align/layout-align.spec.ts
+++ b/src/lib/flex/layout-align/layout-align.spec.ts
@@ -203,6 +203,17 @@ describe('layout-align directive', () => {
               'max-width': '100%'
             }, styler);
       });
+      it('should not add special styles for cross-axis `stretch` when the cross-axis is not `stretch`', () => {
+        createTestComponent(`
+          <div fxLayout
+               fxLayoutAlign='center center'>
+          </div>
+        `);
+  
+        expectNativeEl(fixture).not.toHaveStyle({
+          'max-height': '100%'
+        }, styler);
+      });
     });
 
     describe('for dynamic inputs', () => {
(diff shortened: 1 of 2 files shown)


CaerusKaru:
Once this passes CI, we can get this merged. Great work!

CaerusKaru:
Just fix the linting errors and this should be good to go :+1:

PR #884, opened 2018-11-11
feat(core): add ability to override style building

### LINKED ISSUE #689: poll: what should be the default behavior for fxFlex?
This issue tracks the possibility of changing the default behavior for the `fxFlex` directive. Currently, the behavior is as follows:

* When the input is an integer, this defaults to a percentage value input for `fxFlex`
* When the input is any other singular value, it's interpreted as the intended `flex-basis` value and results in `flex: 1 1 <input>` with `min-width` and `max-width` applied (assuming row format)
* When the input is three values, it's interpreted as a direct translation to the CSS `flex` property, with `min/max-width` applied according to the algorithm specified [here](https://github.com/angular/flex-layout/wiki/fxFlex-API#additional-fxflex-stylings)
* When the input is longer than three values, it is truncated to the first three values

I'll list some alternative options for `fxFlex` behavior below, please vote for your preferred option, or feel free to add your own. If you like the current behavior, please do not add comments, but simply vote for this comment.

CaerusKaru:
This change decouples the style-generation from the actual directives, meaning that the library is now composed of the following:

* Directives that respond to inputs and `matchMedia` events
* Style generation providers that return styles when triggered by directives

This allows for end-users or library authors to provide their own unique style generation (even by borrowing or extending our existing library code) for any directive. This is entirely non-mandatory for use of the `BaseDirective`, since the `BaseDirective` need not always use a de-coupled style provider to function.

The canonical example is the following:
```ts
@Injectable()
export class CustomStyleBuilder extends StyleBuilder {
  buildStyles(input: string) {
	return {
		'style1': 'value1',
	};
  }
}

@NgModule({
	...
	providers: [
		provide: <the style builder to orverride, e.g. FlexStyleBuilder>,
		useClass: CustomStyleBuilder,
	],
})
export class MyAppModule {}
```

Fixes #689

### DIFF
--- a/src/lib/core/base/base.ts
+++ b/src/lib/core/base/base.ts
@@ -21,6 +21,7 @@ import {
 import {ResponsiveActivation, KeyOptions} from '../responsive-activation/responsive-activation';
 import {MediaMonitor} from '../media-monitor/media-monitor';
 import {MediaQuerySubscriber} from '../media-change';
+import {StyleBuilder} from '../style-builder/style-builder';
 
 /** Abstract base class for the Layout API styling directives. */
 export abstract class BaseDirective implements OnDestroy, OnChanges {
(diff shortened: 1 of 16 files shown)


CaerusKaru:
Needs rebase before merge

PR #888, opened 2018-11-14
feat(core): add memoization to style generation

### LINKED ISSUE #887: feat(core): add memoization for all style generation

CaerusKaru:
Fixes #887

### DIFF
--- a/src/lib/core/base/base.ts
+++ b/src/lib/core/base/base.ts
@@ -109,9 +109,21 @@ export abstract class BaseDirective implements OnDestroy, OnChanges {
     return this._elementRef.nativeElement;
   }
 
+  /** Add styles to the element using predefined style builder */
   protected addStyles(input: string, parent?: Object) {
-    const styles: StyleDefinition = this._styleBuilder!.buildStyles(input, parent);
-    this._applyStyleToElement(styles);
+    let styleDefinition: StyleDefinition;
+    if (this._styleCache.has(input)) {
+      styleDefinition = this._styleCache.get(input)!;
+      this._applyStyleToElement(styleDefinition);
+    } else {
+      const {shouldCache, styles} = this._styleBuilder!.buildStyles(input, parent);
+      this._applyStyleToElement(styles);
+      if (shouldCache) {
+        this._styleCache.set(input, styles);
+      }
+      styleDefinition = styles;
+    }
+    this._styleBuilder!.sideEffect(input, styleDefinition, parent);
   }
 
   /** Access the current value (if any) of the @Input property */
@@ -246,4 +258,7 @@ export abstract class BaseDirective implements OnDestroy, OnChanges {
    * getComputedStyle() during ngOnInit().
    */
   protected _hasInitialized = false;
+
+  /** Cache map for style computation */
+  protected _styleCache: Map<string, StyleDefinition> = new Map();
 }
--- a/src/lib/core/base/base.ts
+++ b/src/lib/core/base/base.ts
@@ -111,18 +111,20 @@ export abstract class BaseDirective implements OnDestroy, OnChanges {
 
   /** Add styles to the element using predefined style builder */
   protected addStyles(input: string, parent?: Object) {
-    let styleDefinition: StyleDefinition;
-    if (this._styleBuilder!.shouldCache && this._styleCache.has(input)) {
-      styleDefinition = this._styleCache.get(input)!;
-      this._applyStyleToElement(styleDefinition);
-    } else {
-      styleDefinition = this._styleBuilder!.buildStyles(input, parent);
-      this._applyStyleToElement(styleDefinition);
-      if (this._styleBuilder!.shouldCache) {
-        this._styleCache.set(input, styleDefinition);
+    const builder = this._styleBuilder!;
+    const useCache = builder.shouldCache;
+
+    let genStyles: StyleDefinition | undefined = this._styleCache.get(input);
+
+    if (!genStyles || !useCache) {
+      genStyles = builder.buildStyles(input, parent);
+      if (useCache) {
+        this._styleCache.set(input, genStyles);
       }
     }
-    this._styleBuilder!.sideEffect(input, styleDefinition, parent);
+
+    this._applyStyleToElement(genStyles);
+    builder.sideEffect(input, genStyles, parent);
   }
 
   /** Access the current value (if any) of the @Input property */
--- a/src/lib/core/base/base.ts
+++ b/src/lib/core/base/base.ts
@@ -112,16 +112,15 @@ export abstract class BaseDirective implements OnDestroy, OnChanges {
   /** Add styles to the element using predefined style builder */
   protected addStyles(input: string, parent?: Object) {
     let styleDefinition: StyleDefinition;
-    if (this._styleCache.has(input)) {
+    if (this._styleBuilder!.shouldCache && this._styleCache.has(input)) {
       styleDefinition = this._styleCache.get(input)!;
       this._applyStyleToElement(styleDefinition);
     } else {
-      const {shouldCache, styles} = this._styleBuilder!.buildStyles(input, parent);
-      this._applyStyleToElement(styles);
-      if (shouldCache) {
-        this._styleCache.set(input, styles);
+      styleDefinition = this._styleBuilder!.buildStyles(input, parent);
+      this._applyStyleToElement(styleDefinition);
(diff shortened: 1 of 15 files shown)


CaerusKaru:
Ok so the follow-up was a little hasty. The `layout` issue was resolved plainly and easily by adding a `sideEffect` parameter to `addStyles`. Basically it allows you to get the computed styles and perform an operation with them after the styles have been added (the logic being that if you need to change style generation itself, it should be done in the builder).

For gap, the story is different because it depends a lot on the directionality of the given moment. So, while caching is done in grid mode, it's not really possible at any other point. Although this is not ideal, it's an unfortunate reality.

CaerusKaru:
Conceptually this raises issues with `fxLayoutGap` because it doesn't actually leave much room for the `StyleBuilder` to be independent. A closer look will have to be taken at that particular directive.

CaerusKaru:
Ok fixed that too because now the `sideEffects` are part of the `StyleBuilder`s. That way there's a clear-cut path for the chain of events: `StyleBuilder.buildStyles` -> `StyleBuilder.sideEffect`. This way the `buildStyles` values can be cached if necessary, and the `sideEffect` computation can happen regardless.

src/lib/core/base/base.ts
```diff
@@ -109,9 +109,21 @@ export abstract class BaseDirective implements OnDestroy, OnChanges {
...
+      this._applyStyleToElement(styleDefinition);
+    } else {
+      const {shouldCache, styles} = this._styleBuilder!.buildStyles(input, parent);
+      this._applyStyleToElement(styles);
+      if (shouldCache) {
+        this._styleCache.set(input, styles);
+      }
+      styleDefinition = styles;
+    }
+    this._styleBuilder!.sideEffect(input, styleDefinition, parent);
```
ThomasBurleson: Why are you using the `!` here ?

src/lib/core/style-builder/style-builder.ts
```diff
@@ -10,5 +10,14 @@ import {StyleDefinition} from '../style-utils/style-utils';
 
 @Injectable()
 export abstract class StyleBuilder {
-  abstract buildStyles(input: string, parent?: Object): StyleDefinition;
+  abstract buildStyles(input: string, parent?: Object): StyleBuilderOutput;
+
+  sideEffect(_input: string, _styles: StyleDefinition, _parent?: Object) {
```
ThomasBurleson: What is the intent of the `sideEffect`; should we document it here with comments?

src/lib/flex/flex-align/flex-align.ts
```diff
@@ -20,29 +20,32 @@ import {
...
 @Injectable({providedIn: 'root'})
-export class FlexAlignStyleBuilder implements StyleBuilder {
-  buildStyles(input: string): StyleDefinition {
-    const css: {[key: string]: string | number} = {};
+export class FlexAlignStyleBuilder extends StyleBuilder {
+  constructor() {
+    super();
+  }
+  buildStyles(input: string): StyleBuilderOutput {
+    const styles: {[key: string]: string | number} = {};
```
ThomasBurleson: Should we not define a type:

```ts
export type CSSStyle {
    [key: string]: string | number
}
```

and then use it

```ts
const styles : CSSStyle = { };
```

src/lib/flex/flex-fill/flex-fill.ts
```diff
@@ -23,9 +23,12 @@ const FLEX_FILL_CSS = {
 };
 
 @Injectable({providedIn: 'root'})
-export class FlexFillStyleBuilder implements StyleBuilder {
-  buildStyles(_input: string): StyleDefinition {
-    return FLEX_FILL_CSS;
+export class FlexFillStyleBuilder extends StyleBuilder {
+  constructor() {
```
ThomasBurleson: This is not necessary.

src/lib/flex/flex-offset/flex-offset.spec.ts
```diff
@@ -223,9 +223,12 @@ describe('flex-offset directive', () => {
 });
 
 @Injectable({providedIn: FlexModule})
-export class MockFlexOffsetStyleBuilder implements StyleBuilder {
+export class MockFlexOffsetStyleBuilder extends StyleBuilder {
+  constructor() {
```
ThomasBurleson: This constructor is not necessary.

src/lib/core/style-builder/style-builder.ts
```diff
@@ -10,5 +10,14 @@ import {StyleDefinition} from '../style-utils/style-utils';
...
+    // This should be a no-op unless an algorithm is provided in a subclass
+  }
+}
+
+export interface StyleBuilderOutput {
+  styles: StyleDefinition;
+  shouldCache: boolean;
```
ThomasBurleson: This should be optional with default to `true` and not be required to be returned.

src/lib/flex/flex-align/flex-align.ts
```diff
@@ -20,29 +20,32 @@ import {
...
+        styles['align-self'] = 'flex-end';
         break;
       default:
-        css['align-self'] = input;
+        styles['align-self'] = input;
         break;
     }
 
-    return css;
+    return {styles, shouldCache: true};
```
ThomasBurleson: This seems wrong and verbose. I would expect the API to simply `return styles`

src/lib/flex/flex-align/flex-align.ts
CaerusKaru: We actually have a type! It's called `StyleDefinition`, and it wasn't used in the legacy code that I pulled. Refactored to use the actual type

src/lib/core/base/base.ts
CaerusKaru: The style builder is an opt-in on the `BaseDirective`, and therefore can be `undefined`. But, to consciously use `addStyles` on `BaseDirective` implies that you are using a style builder, so the `!` means "trust me, it's there"

src/lib/core/style-builder/style-builder.ts
CaerusKaru: The intent is to have some routine execute regardless of caching, with the result of the computed styles, but totally opt-in. I'll add a small comment for it

src/lib/flex/flex-align/flex-align.ts
CaerusKaru: I've moved `shouldCache` as a member on the `StyleBuilder`, so that it can be dynamically set during style generation (or not).

src/lib/core/base/base.ts
```diff
@@ -109,9 +109,20 @@ export abstract class BaseDirective implements OnDestroy, OnChanges {
     return this._elementRef.nativeElement;
   }
 
+  /** Add styles to the element using predefined style builder */
   protected addStyles(input: string, parent?: Object) {
-    const styles: StyleDefinition = this._styleBuilder!.buildStyles(input, parent);
-    this._applyStyleToElement(styles);
+    let styleDefinition: StyleDefinition;
```
ThomasBurleson: I think this is a more clear version:

```ts
protected addStyles(input: string, parent?: Object) {
  const builder = this.styleBuilder!;
  const useCache = builder.shouldCache;

  let genStyles: StyleDefinition = this._styleCache.get(input);

  if (!genStyles || !useCache) {
    genStyles = builder.buildStyles(input, parent);
    if (useCache) {
      this._styleCache.set(input, genStyles);
    }
  }

  this._applyStyleToElement(genStyles);
  builder.sideEffect(input, genStyles, parent);
}
```

src/lib/core/style-builder/style-builder.ts
```diff
@@ -5,10 +5,22 @@
...
+
+  /** Build the styles given an input string and configuration object from a host */
   abstract buildStyles(input: string, parent?: Object): StyleDefinition;
+
+  /**
+   * Run a side effect computation given the input string and the computed styles
+   * from the build task and the host configuration object
+   * NOTE: This should be a no-op unless an algorithm is provided in a subclass
+   */
+  sideEffect(_input: string, _styles: StyleDefinition, _parent?: Object) {
```
ThomasBurleson: This is still not clear. Can you provide a link or example of where a sideEffect is used?
Is this essential a `notify()` operation?

CaerusKaru: Check out the file for layout-gap. It needs to run a side effect to apply all the correct styles to the children (since `buildStyles` is synonymous with building styles only on the host)

PR #892, opened 2018-11-17
refactor: migrate ObservableMedia to MediaObserver

### LINKED ISSUE #885: GENERAL NOTICE: changes to ObservableMedia
Due to the impending changes to RxJS in v7, there will be some changes made to `ObservableMedia`. This is because in the next version, RxJS no longer allows extending of `Observable` (since it's being changed to a function instead of a class).

One option we will explore is simply migrating to the CDK's `BreakpointObserver`. Another will be wrapping the observable in another class (possibly still called `ObservableMedia` just with a different way of accessing the `Observable`).

This notice is as much open for discussion as it is for getting the news out there. RxJS v7 is still a ways away, but the earlier we plan and explore options, the better the user experience for everyone.

CaerusKaru:
BREAKING CHANGE:
* `ObservableMedia` is now deprecated in anticipation of RxJS v7. The new API is called `MediaObserver`, and provides the exact same functionality as `ObservableMedia`, except you cannot directly subscribe to it
* You can subscribe to `MediaObserver`'s `media$` property in place of subscribing directly to `ObservableMedia`

Fixes #885

### DIFF
--- a/src/lib/core/base/base.ts
+++ b/src/lib/core/base/base.ts
@@ -109,9 +109,21 @@ export abstract class BaseDirective implements OnDestroy, OnChanges {
     return this._elementRef.nativeElement;
   }
 
+  /** Add styles to the element using predefined style builder */
   protected addStyles(input: string, parent?: Object) {
-    const styles: StyleDefinition = this._styleBuilder!.buildStyles(input, parent);
-    this._applyStyleToElement(styles);
+    let styleDefinition: StyleDefinition;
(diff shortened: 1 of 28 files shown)


CaerusKaru:
This is blocked (only technically) on #888

ThomasBurleson:
@CaerusKaru - can you rebase this?

CaerusKaru:
@ThomasBurleson done.

PR #900, opened 2018-12-04
feat(core): add centralized media marshal service

### LINKED ISSUE #692: fxLayoutGap and fxFlexOffset don't work together
It seems that only you can have either Gap or offset but not both. See reproducer for details:

https://stackblitz.com/edit/angular-material-flex-layout-seed-tqyekw

### LINKED ISSUE #903: [Custom Breakpoints] INTENT TO IMPLEMENT: new centralized media marshaller
Historically, Custom Breakpoints has been a sore spot for Angular Layout. It was difficult and confusing to configure, and in the end didn't always work. We have an entire tag now on the issue tracker dedicated just to custom breakpoint issues. So it's time for a new approach.

Previously, to add custom breakpoints you would have to do two things:
1. Create the breakpoints and register them with the library module
2. Create individual directives for each feature you wanted to extend

Unfortunately, due to the structure of our first-party directives, this wasn't easy, leads to performance drag, collisions, and bloat. This was because by extending our first-party directives, you essentially register two of the same directive, and one with a little bit extra. This is very much not ideal, but was a necessary evil due to the structure of our own implementation.

Our directives were implemented in such a way that each directive was responsible for listening to media queries and responding to them by processing their own inputs. Without extending first-party directives, you risk having two competing directives, each with a different idea of what breakpoint should be active at a given time. When you _do_ extend those directives, you get the aforementioned bloat and sometimes a collision anyway between the two directives, depending on which processed the media query last.

So what's the solution? In short, the media query processing should be moved out of the individual directives. In the most straightforward structure, directives should be organized as follows:

1. The base directive, in charge of all common processing and utility methods
2. The selector directive (e.g. `FlexDirective`), in charge of all common processing _for flex inputs_
3. The breakpoint directive, in charge of registering and updating based on inputs that correspond to breakpoints

These would pair with a new centralized MediaQuery processor called the `MediaMarshaller`, which registers new breakpoints from directives, and then fires events to those directives when the breakpoint is triggered. Plain and simple.

This design is implemented in #900 for the flex directives, and will be extended to the grid and extended directives once the design is validated.

It is our hope that this brings the custom breakpoints story to a close for our users, providing a more streamlined API with which to work. An example implementation of a custom breakpoint directive in the new design is as follows (e.g. for `flex-offset`):
```ts
import {Directive} from '@angular/core';
import {FlexOffsetDirective} from '@angular/flex-layout';

const selector = `[fxFlexOffset.xxs], [fxFlexOffset.xxl]`;
const inputs = ['fxFlexOffset.xxs', 'fxFlexOffset.xxl'];

@Directive({selector, inputs})
export class ExtendFlexOffsetDirective extends FlexOffsetDirective {
	protected inputs = inputs;
}
```

**EDIT**: I thought this went without saying, but we would value any and all feedback on this design _before_ we ship it!

CaerusKaru:
### Summary
This design change marks a number of departures from the current Angular Layout configuration:

1. A number of APIs are deprecated in favor of a more polished, integrated API
2. A new semantics for creating custom breakpoint directives is introduced
3. A number of bugs caught along the way and design changes missed in past PRs are rectified

>  **PLEASE NOTE**: There will be **no end-user API changes** as a result of this PR. 

Unless custom breakpoints are configured, devs should see no change in how the library usage. These changes will deliver notice significant size & performance improvements.

### New APIs

* `MediaMarshaller`
	A centralized data store for elements, breakpoints, and key-value pairs. The `MediaMarshaller` responds to mediaQuery changes and triggers appropriate Layout directives. This class also introduces a way to track changes for arbitrary elements and directive types
* `BaseDirective2`
	A new directive with stripped-down functionality from the old `BaseDirective` that is designed to work in symbiosis with the `MediaMarshaller`

### Custom Breakpoints

The custom breakpoints story has changed significantly. Instead of extending directives that contain the default breakpoints and writing lengthy constructors, the process has been paired down to the following (i.e. for `fxFlexOffset`):

```ts
const inputs = ['fxFlexOffset.xss'];
const selector = `[fxFlexOffset.xss]`;

@Directive({selector, inputs})
export class XssFlexOffsetDirective extends FlexOffsetDirective {
	protected inputs = inputs;
}
```

Never again will a change in the base directive constructor require an entire rewrite of custom breakpoint directives. And registering a new directive no longer brings collisions and double-registrations competing with the default directives. Everything is separated and improved.

### Features

* All of the Grid and extended Flex directives have been updated to the new standards; namely, they have been refactored to the new API with `StyleBuilder`s. The notable exception is `show-hide`, which uses a `StyleBuilder`, but does not have a cache for the results. 
> FxShowHide does not use a stylebuilder cache as to avoid complexity for cache-shifting based on the host `display` property. If needed, an internal cache can be easily added later.

### Bug Fixes

* A number of APIs had the default values calculated in the directives instead of the `StyleBuilder`, meaning that it would be much harder to override by end-users. This has been fixed
* An issue where `fxLayoutGap` was canceling out `fxFlexOffset` has been corrected

### Deprecated APIs

* `MediaMonitor` (use `MediaMarshaller`)
* `ResponsiveActivation`
* `BreakpointX`
* `KeyOptions`
* `negativeOf`
* `BaseDirective` (use `BaseDirective2`)
* `BaseAdapter`

### Build Improvements

The minified size of the Angular Layout library has decreased *~38%* from **132KB** to **82KB**, a total savings of 50KB.  After the deprecated APIs are deleted (Beta.21), the build size will be reduced yet again. 

### Performance Improvements

It should also bring about performance improvements as a result of fewer RxJS subscriptions, memoized style lookups, and other API processing. 

### Stabilization 

It is our hope that along with the added `StyleBuilder` functionality, and migration away from `ObservableMedia`, this represents the end-stage towards stability for Angular Layout.

Closes #903 
Fixes #692

### DIFF
--- a/src/apps/demo-app/src/app/github-issues/split/split.directive.ts
+++ b/src/apps/demo-app/src/app/github-issues/split/split.directive.ts
@@ -54,7 +54,7 @@ export class SplitDirective implements AfterContentInit, OnDestroy {
       const currentValue = flex.activatedValue;
 
       // Update Flex-Layout value to build/inject new flexbox CSS
-      flex.activatedValue = this.calculateSize(currentValue, delta);
+      flex.activatedValue = `${this.calculateSize(currentValue, delta)}`;
     });
   }
 
--- a/src/apps/universal-app/src/app/split/split.directive.ts
+++ b/src/apps/universal-app/src/app/split/split.directive.ts
@@ -60,7 +60,7 @@ export class SplitDirective implements AfterContentInit, OnDestroy {
       const currentValue = flex.activatedValue;
 
       // Update Flex-Layout value to build/inject new flexbox CSS
-      flex.activatedValue = this.calculateSize(currentValue, delta);
+      flex.activatedValue = `${this.calculateSize(currentValue, delta)}`;
     });
   }
 
--- a/src/lib/core/base/index.ts
+++ b/src/lib/core/base/index.ts
@@ -8,3 +8,4 @@
 
 export * from './base';
 export * from './base-adapter';
+export * from './base2';
--- a/src/lib/core/breakpoints/break-point.ts
+++ b/src/lib/core/breakpoints/break-point.ts
@@ -10,4 +10,6 @@ export interface BreakPoint {
   alias: string;
   suffix?: string;
   overlapping?: boolean;
+  // The priority of the individual breakpoint when overlapping another breakpoint
+  priority?: number;
 }
--- a/src/lib/core/media-marshaller/media-marshaller.ts
+++ b/src/lib/core/media-marshaller/media-marshaller.ts
@@ -1,0 +1,227 @@
+/**
+ * @license
+ * Copyright Google LLC All Rights Reserved.
+ *
+ * Use of this source code is governed by an MIT-style license that can be
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {Injectable} from '@angular/core';
+import {merge, Observable, Subject, Subscription} from 'rxjs';
+import {filter} from 'rxjs/operators';
+
+import {BreakPoint} from '../breakpoints/break-point';
+import {prioritySort} from '../breakpoints/breakpoint-tools';
+import {BreakPointRegistry} from '../breakpoints/break-point-registry';
+import {MatchMedia} from '../match-media/match-media';
+import {MediaChange} from '../media-change';
+
+type Builder = Function;
+type ValueMap = Map<string, string>;
+type BreakpointMap = Map<string, ValueMap>;
+type ElementMap = Map<HTMLElement, BreakpointMap>;
+type SubscriptionMap = Map<string, Subscription>;
+type WatcherMap = WeakMap<HTMLElement, SubscriptionMap>;
+type BuilderMap = WeakMap<HTMLElement, Map<string, Builder>>;
+
+export interface ElementMatcher {
+  element: HTMLElement;
+  key: string;
+  value: any;
+}
+
+/**
+ * MediaMarshaller - register responsive values from directives and
+ *                   trigger them based on media query events
+ */
+@Injectable({providedIn: 'root'})
+export class MediaMarshaller {
+  private activatedBreakpoints: BreakPoint[] = [];
+  private elementMap: ElementMap = new Map();
+  private watcherMap: WatcherMap = new WeakMap();
+  private builderMap: BuilderMap = new WeakMap();
+  private subject: Subject<ElementMatcher> = new Subject();
+
+  get activatedBreakpoint(): string {
+    return this.activatedBreakpoints[0] ? this.activatedBreakpoints[0].alias : '';
+  }
+
+  constructor(protected matchMedia: MatchMedia,
+              protected breakpoints: BreakPointRegistry) {
+    this.matchMedia.observe().subscribe(this.activate.bind(this));
+  }
+
+  /**
+   * activate or deactivate a given breakpoint
+   * @param mc
+   */
+  activate(mc: MediaChange) {
+    const bp: BreakPoint | null = this.findByQuery(mc.mediaQuery);
+    if (mc.matches && bp) {
+      this.activatedBreakpoints.push(bp);
+      this.activatedBreakpoints.sort(prioritySort);
+    } else if (!mc.matches && bp) {
+      // Remove the breakpoint when it's deactivated
+      this.activatedBreakpoints.splice(this.activatedBreakpoints.indexOf(bp), 1);
+    }
+    this.updateStyles();
+  }
+
+  /**
+   * initialize the marshaller with necessary elements for delegation on an element
+   * @param element
+   * @param key
+   * @param builder optional so that custom bp directives don't have to re-provide this
+   * @param observables
+   */
+  init(element: HTMLElement,
+       key: string,
+       builder?: Builder,
+       observables: Observable<any>[] = []): void {
+    if (builder) {
+      let builders = this.builderMap.get(element);
+      if (!builders) {
+        builders = new Map();
+        this.builderMap.set(element, builders);
+      }
+      builders.set(key, builder);
+    }
+    if (observables) {
+      let watchers = this.watcherMap.get(element);
+      if (!watchers) {
+        watchers = new Map();
+        this.watcherMap.set(element, watchers);
+      }
+      const subscription = watchers.get(key);
+      if (!subscription) {
+        const newSubscription = merge(...observables).subscribe(() => {
+          const currentValue = this.getValue(element, key);
+          this.updateElement(element, key, currentValue);
+        });
+        watchers.set(key, newSubscription);
+      }
+    }
+  }
+
+  /**
+   * get the value for an element and key and optionally a given breakpoint
+   * @param element
+   * @param key
+   * @param bp
+   */
+  getValue(element: HTMLElement, key: string, bp?: string): any {
+    const bpMap = this.elementMap.get(element);
+    if (bpMap) {
+      const values = bp !== undefined ? bpMap.get(bp) : this.getFallback(bpMap);
+      if (values) {
+        const value = values.get(key);
+        return value !== undefined ? value : '';
+      }
+    }
+    return '';
+  }
+
+  /**
+   * whether the element has values for a given key
+   * @param element
+   * @param key
+   */
+  hasValue(element: HTMLElement, key: string): boolean {
+    const bpMap = this.elementMap.get(element);
+    if (bpMap) {
+      const values = this.getFallback(bpMap);
+      if (values) {
+        return values.get(key) !== undefined || false;
+      }
+    }
+    return false;
+  }
+
+  /**
+   * Set the value for an input on a directive
+   * @param element the element in question
+   * @param key the type of the directive (e.g. flex, layout-gap, etc)
+   * @param bp the breakpoint suffix (empty string = default)
+   * @param val the value for the breakpoint
+   */
+  setValue(element: HTMLElement, key: string, val: any, bp: string): void {
+    let bpMap: BreakpointMap | undefined = this.elementMap.get(element);
+    if (!bpMap) {
+      bpMap = new Map().set(bp, new Map().set(key, val));
+      this.elementMap.set(element, bpMap);
+    } else {
+      const values = (bpMap.get(bp) || new Map()).set(key, val);
+      bpMap.set(bp, values);
+      this.elementMap.set(element, bpMap);
+    }
+    this.updateElement(element, key, this.getValue(element, key));
+  }
+
+  trackValue(element: HTMLElement, key: string): Observable<ElementMatcher> {
+    return this.subject.asObservable()
+      .pipe(filter(v => v.element === element && v.key === key));
+  }
+
+  /** update all styles for all elements on the current breakpoint */
+  updateStyles(): void {
+    this.elementMap.forEach((bpMap, el) => {
+      const valueMap = this.getFallback(bpMap);
+      if (valueMap) {
+        valueMap.forEach((v, k) => this.updateElement(el, k, v));
+      }
+    });
+  }
+
+  /**
+   * update a given element with the activated values for a given key
+   * @param element
+   * @param key
+   * @param value
+   */
+  updateElement(element: HTMLElement, key: string, value: any): void {
+    const builders = this.builderMap.get(element);
+    if (builders) {
+      const builder: Builder | undefined = builders.get(key);
+      if (builder) {
+        builder(value);
+        this.subject.next({element, key, value});
+      }
+    }
+  }
+
+  /**
+   * release all references to a given element
+   * @param element
+   */
+  releaseElement(element: HTMLElement): void {
+    const watcherMap = this.watcherMap.get(element);
+    if (watcherMap) {
+      watcherMap.forEach(s => s.unsubscribe());
+      this.watcherMap.delete(element);
+    }
+    const elementMap = this.elementMap.get(element);
+    if (elementMap) {
+      elementMap.forEach((_, s) => elementMap.delete(s));
+      this.elementMap.delete(element);
+    }
+  }
+
+  /** Breakpoint locator by mediaQuery */
+  private findByQuery(query: string) {
+    return this.breakpoints.findByQuery(query);
+  }
+
+  /**
+   * get the fallback breakpoint for a given element, starting with the current breakpoint
+   * @param bpMap
+   */
+  private getFallback(bpMap: BreakpointMap): ValueMap | undefined {
+    for (let i = 0; i < this.activatedBreakpoints.length; i++) {
+      const activatedBp = this.activatedBreakpoints[i];
+      const valueMap = bpMap.get(activatedBp.alias);
+      if (valueMap) {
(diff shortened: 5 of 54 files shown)


src/lib/core/breakpoints/break-point.ts
```diff
@@ -10,4 +10,6 @@ export interface BreakPoint {
   alias: string;
   suffix?: string;
   overlapping?: boolean;
+  priority?: number;
+  includes?: string[];
```
CaerusKaru: To be removed, but a possible design for breakpoint prioritization

src/lib/core/media-marshaller/media-marshaller.ts
```diff
@@ -0,0 +1,175 @@
...
+        builder(val);
+        const parentSubject = this.parentSubjects.get(element);
+        if (parentSubject) {
+          parentSubject.next();
+        }
+      }
+    }
+  }
+
+  destroy(): void {
```
CaerusKaru: This needs to be fleshed out due to the subscribe-y nature of this, but also the ephemeral nature of services.

CaerusKaru:
Still needs the following after design sign-off:
* Unit tests for `MediaMarshaller`
* Unit tests for `NewBaseDirective`
* New name and deprecation strategy for `NewBaseDirective` and `BaseDirective`
* Move `ngOnChanges` into `NewBaseDirective`?

src/lib/flex/flex-align/flex-align.ts
```diff
@@ -96,33 +83,23 @@ export class FlexAlignDirective extends BaseDirective implements OnInit, OnChang
...
-  /**
-   * After the initial onChanges, build an mqActivation object that bridges
-   * mql change events to onMediaQueryChange handlers
-   */
-  ngOnInit() {
-    super.ngOnInit();
-
-    this._listenForMediaQueryChanges('align', 'stretch', (changes: MediaChange) => {
-      this._updateWithValue(changes.value);
+    // TODO: figure out how custom breakpoints interact with this method
```
charsleysa: Could change the inputs to be a protected class member allowing custom directives which override this directive to specify the custom breakpoints. That way this directive handles the default breakpoints and the custom directive handles the custom breakpoints, ngOnChanges won't be called for each others inputs only their own inputs.

CaerusKaru: That's exactly what I had in mind 😄

CaerusKaru:
NOTE: there is a critical issue with caching and directionality right now affecting `fxFlex`

CaerusKaru:
Issue has been resolved, this PR is now ready for review :+1:

src/lib/core/base/new-base.ts
```diff
@@ -0,0 +1,114 @@
...
+ * found in the LICENSE file at https://angular.io/license
+ */
+import {ElementRef, OnChanges, SimpleChanges} from '@angular/core';
+
+import {StyleDefinition, StyleUtils} from '../style-utils/style-utils';
+import {StyleBuilder} from '../style-builder/style-builder';
+import {MediaMarshaller} from '../media-marshaller/media-marshaller';
+import {buildLayoutCSS} from '../../utils/layout-validator';
+
+export abstract class NewBaseDirective implements OnChanges {
```
ThomasBurleson: Similar to Renderer2, let's call this `BaseDirective2` instead of NewBaseDirective

src/apps/demo-app/src/app/github-issues/split/split.directive.ts
```diff
@@ -54,7 +54,7 @@ export class SplitDirective implements AfterContentInit, OnDestroy {
       const currentValue = flex.activatedValue;
 
       // Update Flex-Layout value to build/inject new flexbox CSS
-      flex.activatedValue = this.calculateSize(currentValue, delta);
+      flex.activatedValue = this.calculateSize(currentValue, delta) + '';
```
ThomasBurleson:
```ts
const currentValue = this.calculateSize(flex.activatedValue, delta);
flex.activatedValue = `${currentValue}`
```

src/apps/universal-app/src/app/split/split.directive.ts
```diff
@@ -60,7 +60,7 @@ export class SplitDirective implements AfterContentInit, OnDestroy {
       const currentValue = flex.activatedValue;
 
       // Update Flex-Layout value to build/inject new flexbox CSS
-      flex.activatedValue = this.calculateSize(currentValue, delta);
+      flex.activatedValue = this.calculateSize(currentValue, delta) + '';
```
ThomasBurleson:
```ts
const currentValue = this.calculateSize(flex.activatedValue, delta);
flex.activatedValue = `${currentValue}`
```

src/lib/core/base/index.ts
```diff
@@ -8,3 +8,4 @@
 
 export * from './base';
 export * from './base-adapter';
+export * from './new-base';
```
ThomasBurleson:
```ts
export * from './base2';
```

src/lib/core/breakpoints/break-point.ts
ThomasBurleson: Let's add comments to document api adds.

src/lib/core/media-marshaller/media-marshaller.ts
```diff
@@ -0,0 +1,218 @@
...
+   * activate or deactivate a given breakpoint
+   * @param mc
+   */
+  activate(mc: MediaChange) {
+    const bp: BreakPoint | null = this.findByQuery(mc.mediaQuery);
+    if (mc.matches && bp) {
+      this.activatedBreakpoints.push(bp);
+      this.activatedBreakpoints.sort(prioritySort);
+    } else if (!mc.matches && bp) {
+      this.activatedBreakpoints.splice(this.activatedBreakpoints.indexOf(bp), 1);
```
ThomasBurleson: Should comment here what is going on... why we are splicing.

src/lib/core/media-marshaller/media-marshaller.ts
```diff
@@ -0,0 +1,218 @@
...
+  /**
+   * initialize the marshaller with necessary elements for delegation on an element
+   * @param element
+   * @param key
+   * @param builder optional so that custom bp directives don't have to re-provide this
+   * @param observables
+   */
+  init(element: HTMLElement,
+       key: string,
+       builder?: Function,
```
ThomasBurleson: Should we type this as a StyleBuilderFactory ?

src/lib/core/media-marshaller/media-marshaller.ts
```diff
@@ -0,0 +1,218 @@
...
+   * update a given element with the activated values for a given key
+   * @param element
+   * @param key
+   * @param val
+   */
+  updateElement(element: HTMLElement, key: string, val: any): void {
+    const builders = this.builderMap.get(element);
+    const subjects = this.subjectMap.get(element);
+    if (builders) {
+      const builder = builders.get(key);
```
ThomasBurleson: For clarity, let's type the variable `builder`

src/lib/core/media-marshaller/media-marshaller.ts
```diff
@@ -0,0 +1,218 @@
...
+        watchers = new Map();
+        this.watcherMap.set(element, watchers);
+      }
+      const subscription = watchers.get(key);
+      if (!subscription) {
+        const newSubscription = merge(...observables).subscribe(() => {
+          const currentValue = this.getValue(element, key);
+          this.updateElement(element, key, currentValue);
+        });
+        watchers.set(key, newSubscription);
```
ThomasBurleson: Where do we unsubscribe when the DOM element is destroyed?

src/lib/core/media-marshaller/media-marshaller.ts
```diff
@@ -0,0 +1,218 @@
...
+
+  trackValue(element: HTMLElement, key: string): Observable<string> {
+    let elementMap = this.subjectMap.get(element);
+    if (!elementMap) {
+      elementMap = new Map();
+      this.subjectMap.set(element, elementMap);
+    }
+    let subject = elementMap.get(key);
+    if (!subject) {
+      subject = new Subject();
```
ThomasBurleson: We could use a single Subject and filter by key/element. Then the single subject becomes the central stream for all events.

src/lib/core/media-marshaller/media-marshaller.ts
ThomasBurleson: We should talk about this one.

src/lib/extended/class/class.ts
```diff
@@ -12,155 +12,80 @@ import {
...
-   * Delegate value changes to the internal `_ngClassInstance` for processing
-   */
-  protected _base: BaseDirectiveAdapter;   // used for `ngClass.xxx` selectors
+/**
+ * Directive to add responsive support for ngClass.
+ * This maintains the core functionality of 'ngClass' and adds responsive API
+ * Note: this class is a no-op when rendered on the server
+ */
+@Directive({selector, inputs})
+export class DefaultClassDirective extends ClassDirective {
```
ThomasBurleson: I assume you are doing this to preserve support for deprecated `ClassDirective` while introduce the new `DefaultClassDirective` ?
If yes, this means that after the deprecated API is deleted, the `DefaultClassDirective` is the only directive?

src/lib/extended/module.ts
```diff
@@ -8,17 +8,17 @@
...
+import {DefaultShowHideDirective} from './show-hide/show-hide';
+import {DefaultStyleDirective} from './style/style';
 
 
 const ALL_DIRECTIVES = [
-  ShowHideDirective,
-  ClassDirective,
-  StyleDirective,
-  ImgSrcDirective
+  DefaultShowHideDirective,
```
ThomasBurleson: If you are not exporting the Deprecated API, then they cannot be used. Is this your intent?

src/lib/extended/show-hide/show.spec.ts
```diff
@@ -348,26 +334,12 @@ describe('show directive', () => {
...
+const selector = `[fxShow.sm-md], [fxHide.sm-md]`;
 
-  @Input('fxShow.sm-md') set showSmMd(val: string) {
-    this._cacheInput('showSmMd', val);
-  }
-  @Input('fxHide.sm-md') set hideSmMd(val: string) {
-    this._cacheInput('showSmMd', negativeOf(val));
-  }
+@Directive({inputs, selector})
+class FxShowHideDirective extends ShowHideDirective {
```
ThomasBurleson: Add comment that this is for testing a custom directive...

src/lib/flex/flex/flex.ts
```diff
@@ -196,136 +186,119 @@ export class FlexStyleBuilder extends StyleBuilder {
...
+  constructor(protected elRef: ElementRef,
+              protected styleUtils: StyleUtils,
+              @Inject(LAYOUT_CONFIG) protected layoutConfig: LayoutConfigOptions,
+              protected styleBuilder: FlexStyleBuilder,
+              protected marshal: MediaMarshaller) {
+    super(elRef, styleBuilder, styleUtils, marshal);
+    this.marshal.init(this.elRef.nativeElement, this.DIRECTIVE_KEY,
+      this.updateStyle.bind(this));
+    if (this.parentElement) {
+      this.layoutWatcher = this.marshal.trackValue(this.parentElement, 'layout')
```
ThomasBurleson: Let's use `untilViewDestroyed` pipe: https://gist.github.com/ThomasBurleson/37653f1f273149e6f6bc07e5b29ea313


```ts
this.marshal
.trackValue(this.parentElement, 'layout')
.pipe( untilViewDestroyed(elRef) )
.subscribe( this.onLayoutChange.bind(this) );
```

ThomasBurleson: Just add that gist code to utils.

src/lib/flex/layout-gap/layout-gap.ts
```diff
@@ -234,20 +184,56 @@ export class LayoutGapDirective extends BaseDirective
...
+   * Quick accessor to the current HTMLElement's `display` style
+   * Note: this allows us to preserve the original style
+   * and optional restore it when the mediaQueries deactivate
+   */
+  protected getDisplayStyle(source: HTMLElement = this.nativeElement): string {
+    const query = 'display';
+    return this.styler.lookupStyle(source, query);
+  }
+
+  protected buildChildObservable(): void {
```
ThomasBurleson: Nice: 👍

src/lib/extended/class/class.ts
CaerusKaru: `ClassDirective` (and other similar non-directives) are not deprecated, and are still being exported. They just aren't Angular `Directive`s anymore. They're used when you want to create a new custom breakpoint directive, like `DefaultClassDirective`.

src/lib/extended/module.ts
CaerusKaru: Because they're not Angular `Directive`s, they don't need to be listed in the exports for the `NgModule`. They are still exported by the library and are still intended for use (they're not deprecated).

src/lib/core/media-marshaller/media-marshaller.ts
CaerusKaru: I think you've covered this in a later comment, will implement `ngOnDestroy` behavior.

src/lib/core/media-marshaller/media-marshaller.ts
CaerusKaru: `Factory` implies that it returns a `StyleBuilder`, which is not the case here. I've aliased `Builder` to `Function`

src/lib/flex/flex/flex.ts
CaerusKaru: I added similar functionality that I think is less verbose. Let me know what you think

ThomasBurleson: Your use of `destroySubject` in the base class is fine IMO. I would add another property however that exposes a single observable 

```ts
protected destroyed$ = this.destroySubject.asObservable()
```

src/lib/core/media-marshaller/media-marshaller.ts
```diff
@@ -0,0 +1,211 @@
...
+type WatcherMap = WeakMap<HTMLElement, SubscriptionMap>;
+type BuilderMap = WeakMap<HTMLElement, Map<string, Builder>>;
+
+export interface SubjectMatcher {
+  element: HTMLElement;
+  key: string;
+  value: any;
+}
+
+/**
```
ThomasBurleson: Need more description of intent and usage here.

src/lib/core/media-marshaller/media-marshaller.ts
```diff
@@ -0,0 +1,213 @@
...
+
+type Builder = Function;
+type ValueMap = Map<string, string>;
+type BreakpointMap = Map<string, ValueMap>;
+type ElementMap = Map<HTMLElement, BreakpointMap>;
+type SubscriptionMap = Map<string, Subscription>;
+type WatcherMap = WeakMap<HTMLElement, SubscriptionMap>;
+type BuilderMap = WeakMap<HTMLElement, Map<string, Builder>>;
+
+export interface SubjectMatcher {
```
ThomasBurleson: Rename to `ElementMatcher` ?

PR #902, opened 2018-12-04
refactor(): value types for directives expecting simple string or string combinations

vmasek:
This will enable input value completion in IDEs when using `fxLayout`, `fxFlexAlign`, `fxShowHide` and `fxLayoutAlign` directives

I used supported values from directives API docs in [wiki](https://github.com/angular/flex-layout/wiki)
and also included some that are not mentioned there
 - `flex-` prefixes for `start` and `end`
 -  `baseline` for `LayoutAlign` which was not mentioned in wiki [fxLayoutAlign-API](https://github.com/angular/flex-layout/wiki/fxLayoutAlign-API)

--- 

Tested in WebStorm with Angular Language service enabled.

fxflexalign

fxlayout

fxshow

fxlayoutalign

fxlayoutalign-start

### DIFF
--- a/src/lib/extended/show-hide/show-hide.ts
+++ b/src/lib/extended/show-hide/show-hide.ts
@@ -35,6 +35,13 @@ import {Subscription} from 'rxjs';
 
 const FALSY = ['false', false, 0];
 
+type ShowHideValuesType
+  = ''
+  | 'true'
+  | 'false'
+  | number
+  ;
+
 /**
  * For fxHide selectors, we invert the 'value'
  * and assign to the equivalent fxShow selector cache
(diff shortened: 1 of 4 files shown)


vmasek:
I signed it!

CaerusKaru:
In principle, I have no issue with this. It wouldn't affect the runtime build size because all of this is typed metadata, and it's great for developer experience.

Unfortunately, with #900, we're getting rid of individual inputs in favor of making the custom breakpoints story easier for developers, which I think is more of a benefit of this change.

If there is a way to get this in without changing the `@Input` type I would be more than happy to discuss/review.

CaerusKaru:
I've also updated the Wiki for `fxLayoutAlign` to reflect the `baseline` input. Thanks for catching that!

vmasek:
@CaerusKaru I see, is it because of 
`export class CustomShowHideDirective extends ShowHideDirective ...` constructor?
https://github.com/angular/flex-layout/wiki/Breakpoints

It's a pitty that the inputs are removed, it now lacks a type safety and using inputs in metadata decorator is a bad practice according to [angular style guide](https://angular.io/guide/styleguide#decorate-input-and-output-properties).

I'd like to have a closer look on the possibilities, but I don't fully understand the motivation and changes in mentioned PR. Would you have a time for some intro?

CaerusKaru:
@vmasek It's not because of the constructor, we're actually refactoring so that you don't need to include one in a custom breakpoint directive as part of #900.

I have two points about type safety and the style guide. For type safety, inputs generally don't require them, especially in our library where most directives accept any type of value and are coerced to the correct type at runtime. We have documentation that provides the options for our users, and while it would definitely be easier to have an inline reference, maybe a better place for that would be the package README (or the module README?). 

As for the style guide, it is a general best practice guide for _end users_, not library authors. Angular Material, for instance, violates many of the style guide core principles, most notably by using the `host` metadata parameter on practically every directive. It makes it more convenient and maintainable for a more complex codebase, at least in some people's opinions including my own.

If we were to not follow this, and revert back to using `Input` decorators, we would need to declare `n` number of `Input`s based on each component. If we simply declare an array, we don't need to declare any `Input`s, and they all get funneled into central processing by `ngOnChanges`. This gives us the flexibility of saying `ngOnChanges` and the `inputs` array corresponds 1:1 with mediaQuery processing, and all `Input`s are strictly for additional functionality (like `fxShrink` and `fxGrow`).

This is the motivation behind #900, to reduce complexity and overhead introduced by having a ton of `Input`s for each directive, as opposed to centralizing the input functionality to one place. Please head over to #903 for some more explanation and discussion on the issue.

vmasek:
Thank you, in that case, I am closing this PR, if I manage to come up with some other solution for this, I will discuss it in new PR/Issue.

PR #905, opened 2018-12-09
build: switch to circleci

CaerusKaru:


CaerusKaru:
@devversion Can you PTAL? I don't know why the ssr tests are failing. For context, they would run on Travis, but they don't run for me locally with the same error.

The publish task fails bc Jeremy has to copy over the access token.

.circleci/config.yml
```diff
@@ -1,37 +1,319 @@
-# Javascript Node CircleCI 2.0 configuration file
-#
-# Check https://circleci.com/docs/2.0/language-javascript/ for more details
-#
+# Configuration file for https://circleci.com/gh/angular/flex-layout
```
devversion: Maybe also note that this is based on Material's configuration?

.circleci/config.yml
```diff
@@ -1,37 +1,319 @@
...
+    key: *cache_key
+
+# Saves the cache for the current Yarn lock file. We store the node modules and the build
+# cache in order to make subsequent builds faster.
+var_6: &save_cache
+  save_cache:
+    key: *cache_key
+    paths:
+      - "node_modules"
+      - "~/dist/releases"
```
devversion: Why would you want to cache the `release` output? The release output should be transferred through Circle's workspace storage.

.circleci/config.yml
```diff
@@ -1,37 +1,319 @@
...
+  # ----------------------------------------------------------------------------
+  # Job that runs the unit tests on Browserstack. The browsers that will be used
+  # to run the unit tests on Browserstack are set in: test/browser-providers.js
+  # ----------------------------------------------------------------------------
+  tests_browserstack:
+    <<: *job_defaults
+    resource_class: xlarge
+    environment:
+      BROWSER_STACK_USERNAME: "angularteam1"
+      BROWSER_STACK_ACCESS_KEY: "CaXMeMHD9pr5PHg8N7Jq"
```
devversion: We should at some point figure out what to do about licenses.. I'm afraid that we all often run into rate limit exceptions and adding more projects might not improve that situation.

.circleci/config.yml
```diff
@@ -1,37 +1,319 @@
...
-      # Download and cache dependencies
-      - restore_cache:
-          keys:
-          - v1-dependencies-{{ checksum "yarn.lock" }}
-          # fallback to using the latest cache if no exact match is found
-          - v1-dependencies-
+  # -------------------------------------------------------------------------
+  # Job that makes sure Angular Layout can be integrated with a basic app
+  # -------------------------------------------------------------------------
+  hw_build:
```
devversion: `hw_build` is pretty unclear at first glance. Please do not abbreviate the job name.

.circleci/config.yml
```diff
@@ -1,37 +1,319 @@
...
+          command: |
+            mkdir -p /tmp/layout-umd-minified-bundles
+            cp dist/releases/flex-layout/bundles/*.umd.min.js /tmp/layout-umd-minified-bundles
+      # Publish bundle artifacts which will be used to calculate the size change.
+      # Note: Make sure that the size plugin from the Angular robot fetches the artifacts
+      # from this CircleCI job (see .github/angular-robot.yml). Additionally any artifacts need to
+      # be stored with the following path format: "{projectName}/{context}/{fileName}"
+      # This format is necessary because otherwise the bot is not able to pick up the
+      # artifacts from CircleCI. See:
+      # https://github.com/angular/github-robot/blob/master/functions/src/plugins/size.ts#L392-L394
```
devversion: Are you actually using the Github robot?

This seems to be out of scope for this PR because there is no bot configuration yet.

.circleci/config.yml
```diff
@@ -1,37 +1,319 @@
...
+# General setup for CircleCI
+# ---------------------------
+general:
+  branches:
+    only:
+      - master
+      # 5.2.x, 6.0.x, etc
+      - /\d+\.\d+\.x/
+      # 5.x, 6.x, etc
+      - /\d+\.x/
```
devversion: Does `flex-layout` use publish branches? Could be probably removed.

.travis.yml
```diff
@@ -16,49 +16,51 @@ branches:
...
-  - ./scripts/ci/travis-script.sh
-
-cache:
-  yarn: true
-  directories:
-    - ./node_modules/
+#cache:
+#  yarn: true
+#  directories:
+#    - ./node_modules/
```
devversion: If this file is mostly just commented code, just remove it. Travis will be gone once that PR has been merged.

test/browser-providers.js
```diff
@@ -2,81 +2,66 @@
 
 /*
  * Browser Configuration for the different jobs in the CI.
- * Target can be either: BS (Browserstack) | SL (Saucelabs) | TC (Travis CI) | null (To not run)
+ *
+ *   - local: Launches the browser locally on the current operating system.
+ *   - BS: Launches the browser within BrowserStack
+ *   - SL: Launches the browser within Saucelabs
+ *
+ * TODO(devversion): rename this to "browserstack" and "saucelabs".
```
devversion: I've already renamed this to `browserstack` and `saucelabs`. I somewhere have a PR on Material that removes this (just not merged yet).

You can remove this todo.

test/karma.conf.js
```diff
@@ -25,18 +23,22 @@ module.exports = (config) => {
...
 
+      // Include all Angular dependencies
       {pattern: 'node_modules/@angular/**/*', included: false, watched: false},
       {pattern: 'node_modules/rxjs/**/*', included: false, watched: false},
 
       {pattern: 'test/karma-system-config.js', included: true, watched: false},
       {pattern: 'test/karma-test-shim.js', included: true, watched: false},
 
+      // Include a Material theme in the test suite.
+      {pattern: 'dist/packages/**/core/theming/prebuilt/indigo-pink.css', included: true, watched: true},
```
devversion: Do you actually use Material in any of your unit tests?

test/karma.conf.js
```diff
@@ -25,18 +23,22 @@ module.exports = (config) => {
       {pattern: 'node_modules/zone.js/dist/jasmine-patch.min.js', included: true, watched: false},
       {pattern: 'node_modules/zone.js/dist/async-test.js', included: true, watched: false},
       {pattern: 'node_modules/zone.js/dist/fake-async-test.js', included: true, watched: false},
+      {pattern: 'node_modules/hammerjs/hammer.min.js', included: true, watched: false},
```
devversion: Same as for Material, do you use hammer as well as moment in unit tests?

test/protractor.conf.js
```diff
@@ -24,37 +24,24 @@ const config = {
       path: '../tools/axe-protractor/axe-protractor.js',
 
       rules: [
-        // Exclude md-menu elements because those are empty if not active.
-        { id: 'aria-required-children', selector: '*:not(md-menu)' },
+        // Exclude mat-menu elements because those are empty if not active.
+        { id: 'aria-required-children', selector: '*:not(mat-menu)' },
```
devversion: You should just remove these rules for flex-layout. There is no test for `mat-menu` AFAIK.

devversion:
@CaerusKaru Before this can be merged, the `tests_ssr` job should definitely be green, right?

cc. @jelbourn for the CircleCI publish token.

.circleci/config.yml
CaerusKaru: Yes definitely. I'll also credit you explicitly because this work is fantastic.

.circleci/config.yml
CaerusKaru: We have our own, and I'll add it here to avoid using up the main one, but I just wanted to validate the format first

.circleci/config.yml
CaerusKaru: Will change to `hello_world_build`

.circleci/config.yml
CaerusKaru: I think we'd like to start, if anything to track size regressions. I figure we could add the config here and then onboard with the robot later?

.travis.yml
CaerusKaru: Travis is still required, so if this gets merged before that is turned off, it still needs to run. If Jeremy turns off Travis pre-merge I'll delete the file, ow will remove in a follow-up

test/karma.conf.js
CaerusKaru: Yes we do! I know it's surprising, though we probably don't need this theme.

.circleci/config.yml
devversion: Cool, good to know!

test/karma.conf.js
CaerusKaru: Most likely not, will remove

test/protractor.conf.js
CaerusKaru: Yeah I think our only Material test dependency is on button. Either way, will remove

.circleci/config.yml
CaerusKaru: Will follow-up on this one offline

CaerusKaru:
@devversion Thanks for the review, SSR tests definitely need to be green for this to get merged. I'll work on it.

.circleci/config.yml
devversion: Yeah that works.

CaerusKaru:
@devversion resolved all comments, PTAL :+1:

CaerusKaru:
Also marking as blocked until SSR issues can be resolved

.circleci/config.yml
```diff
@@ -1,37 +1,320 @@
...
+  # ----------------------------------------------------------------------------
+  # Job that runs the unit tests on Saucelabs. The browsers that will be used
+  # to run the unit tests on Saucelabs are set in: test/browser-providers.js
+  # ----------------------------------------------------------------------------
+  tests_saucelabs:
+    <<: *job_defaults
+    resource_class: xlarge
+    environment:
+      SAUCE_USERNAME: "angular-ci"
+      SAUCE_ACCESS_KEY: "9b988f434ff8-fbca-8aa4-4ae3-35442987"
```
devversion: What about that license? Nice in regards to BrowserStack

.circleci/config.yml
```diff
@@ -1,37 +1,320 @@
...
+
+      - run: ./scripts/circleci/run-browserstack-tests.sh
+
+  # ----------------------------------------------------------------------------
+  # Job that runs the unit tests on Saucelabs. The browsers that will be used
+  # to run the unit tests on Saucelabs are set in: test/browser-providers.js
+  # ----------------------------------------------------------------------------
+  tests_saucelabs:
+    <<: *job_defaults
+    resource_class: xlarge
```
devversion: Flex-layout should probably go down with the resource class here. On Material we have it enabled for tests because we have around 3k. `Medium` is the default and should be good enough, thoughts?

(same applies for browserstack, local browsers and saucelabs)

.circleci/config.yml
CaerusKaru: Discussed offline, will not switch for SauceLabs at the moment. Could revisit in the future

.circleci/config.yml
CaerusKaru: I kind of like that it runs super-quick though 😉 . But you're probably right.  The default should probably be plenty. I'll try it out in a fixup commit, and if it's not good enough I'll revert back

PR #918, opened 2018-12-15
fix(fxFlexOrder): remove order after responsive deactivation

### LINKED ISSUE #296: feat(breakpoints): only apply flex properties if a breakpoint is activated

### LINKED ISSUE #697: fxLayoutGap not working with fxLayout fxFlexOrder
#### Bug:
fxLayoutGap does not work when fxFlexOrder is given with responsive breakpoints.Initially it is rendered properly. But when the screen is resized,fxLayoutGap does not work.

fxLayoutGap should work when screen is resized.

[stackblitz](https://stackblitz.com/edit/angular-flex-layout-seed-xseaux?file=app/test.component.ts)

1. Initial Page rendered.

image

2. After page resize
image

3. normal mode after resizing.

image

CaerusKaru:
Fixes #697 
Fixes #296

### DIFF
--- a/src/lib/core/base/base2.ts
+++ b/src/lib/core/base/base2.ts
@@ -6,7 +6,7 @@
  * found in the LICENSE file at https://angular.io/license
  */
 import {ElementRef, OnChanges, OnDestroy, SimpleChanges} from '@angular/core';
-import {Subject} from 'rxjs';
+import {Observable, Subject} from 'rxjs';
 
 import {StyleDefinition, StyleUtils} from '../style-utils/style-utils';
 import {StyleBuilder} from '../style-builder/style-builder';
@@ -18,6 +18,9 @@ export abstract class BaseDirective2 implements OnChanges, OnDestroy {
   protected DIRECTIVE_KEY = '';
   protected inputs: string[] = [];
   protected destroySubject: Subject<void> = new Subject();
+  protected observables: Observable<any>[] = [];
+  /** The least recently used styles for the builder */
+  protected lru: StyleDefinition = {};
 
   /** Access to host element's parent DOM node */
   protected get parentElement(): HTMLElement | null {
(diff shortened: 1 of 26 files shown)


src/lib/core/base/base2.ts
```diff
@@ -18,6 +18,9 @@ export abstract class BaseDirective2 implements OnChanges, OnDestroy {
   protected DIRECTIVE_KEY = '';
   protected inputs: string[] = [];
   protected destroySubject: Subject<void> = new Subject();
+  protected observables: Observable<any>[] = [];
```
ThomasBurleson: Can we find a better name for this?

src/lib/core/base/base2.ts
```diff
@@ -78,10 +86,20 @@ export abstract class BaseDirective2 implements OnChanges, OnDestroy {
       }
     }
 
+    this.mru = {...genStyles};
     this.applyStyleToElement(genStyles);
     builder.sideEffect(input, genStyles, parent);
   }
 
+  /** Remove generated styles from an element using predefined style builder */
+  protected clearStyles() {
```
ThomasBurleson: Under what conditions should `clearStyles()` be called?

CaerusKaru: When there isn't fallback value for a given element and key, it indicates that there is no actual value for the activated breakpoints, meaning no styles should be applied. To accomplish this, we clear out the most recently used styles by calling this method.

src/lib/core/base/base2.ts
CaerusKaru: How about `triggerObservables`?

ThomasBurleson:
closed for https://github.com/angular/flex-layout/pull/926

PR #929, opened 2018-12-17
fix(flex-offset): correct type when using with 'strictNullChecks'

Teamop:
Currently, when using `@angular/flex-layout@latest`(7.0.0-beta.21) with `"strictNullChecks": true`, got error: 
image
If with `"strictNullChecks": false`, got no error, so here just correct the return type of `buildStyles`.

@CaerusKaru @ThomasBurleson


Teamop:
Don't know why the checks don't run...

Teamop:
@CaerusKaru done

CaerusKaru:
@Teamop thank you for patching this!

iNViTiON:
Thank you

PR #931, opened 2018-12-17
fix(fxLayoutGap): account for responsive fxHide on children elements

### LINKED ISSUE #606: fxLayoutGap not allows working on resize
If you load the page with an element hidden then resize to where it is visible the gap is not calculated (leaving no space).  Where if the page is loaded where it is visible you can resize all you want and the gap displays correctly. 

*  https://stackblitz.com/edit/angular-flex-layout-seed-v5tsns?file=app/test.component.ts

So it appears that the gap is only calculated on init of the page, and not on the change event. This might have been done for performance ... but leaves for inconstant ui.
Or it was just overlooked. 

Angular: 5.2.0
flex-layout: 2.0.0-beta.12

CaerusKaru:
Fixes #606

### DIFF
--- a/src/lib/flex/layout-gap/layout-gap.ts
+++ b/src/lib/flex/layout-gap/layout-gap.ts
@@ -173,7 +173,7 @@ export class LayoutGapDirective extends BaseDirective2 implements AfterContentIn
     }
     // Gather all non-hidden Element nodes
     const items = this.childrenNodes
-      .filter(el => el.nodeType === 1 && this.getDisplayStyle(el) !== 'none')
+      .filter(el => el.nodeType === 1 && this.willDisplay(el))
       .sort((a, b) => {
         const orderA = +this.styler.lookupStyle(a, 'order');
         const orderB = +this.styler.lookupStyle(b, 'order');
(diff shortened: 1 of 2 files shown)


src/lib/flex/layout-gap/layout-gap.ts
```diff
@@ -200,14 +200,11 @@ export class LayoutGapDirective extends BaseDirective2 implements AfterContentIn
...
-   * Quick accessor to the current HTMLElement's `display` style
-   * Note: this allows us to preserve the original style
-   * and optional restore it when the mediaQueries deactivate
-   */
-  protected getDisplayStyle(source: HTMLElement = this.nativeElement): string {
-    const query = 'display';
-    return this.styler.lookupStyle(source, query);
+  /** Determine if an element will show or hide based on current activation */
+  protected willDisplay(source: HTMLElement): boolean {
+    const value = this.marshal.getValue(source, 'show-hide');
```
ThomasBurleson: How about this for better clarity:

```ts
const value = this.marshal.getValue(source, 'show-hide');
if ( value == '' || value == undefined) {
  value = this.styleUtils.lookupStyle(source, 'display') !== 'none';
}
return value;
```

CaerusKaru: I don't think the value gain is great in terms of clarity: the value returned from the marshaller will never be undefined.

CaerusKaru:
Blocked because this doesn't entirely fix the issue. Will investigate further today.

PR #939, opened 2018-12-18
fix(ngStyle): should work with preexisting styles

### LINKED ISSUE #936: style attribute is not applied anymore when using ngStyle (version 7.x)
# Bug Report

In the following div the style attribute should work...

```

<div style="background-color: red; height: 100px; width: 100px;" [ngStyle]="this.divStyle">
  First div
</div>


```

When I import FlexLayoutModule with @angular/flex-layout - version 7.x the style attribute does not work as expected.. This means it is ignored and only the settings from ngStyle are used.

When I use version 6.x of @angular/flex-layout the style attribute works as expected

https://stackblitz.com/edit/angular-cavljx - shows the non working version - updating the angular-flexlayout version to 6 shows the working version

angular 7.1.3, @angular/flex-layout 7.0.0-beta.21

CaerusKaru:
Fixes #936

### DIFF
--- a/src/lib/extended/style/style.ts
+++ b/src/lib/extended/style/style.ts
@@ -35,6 +35,7 @@ import {
 export class StyleDirective extends BaseDirective2 implements DoCheck {
 
   protected DIRECTIVE_KEY = 'ngStyle';
+  protected fallbackStyles: NgStyleMap = {};
 
   constructor(protected elementRef: ElementRef,
               protected styler: StyleUtils,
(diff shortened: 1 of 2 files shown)


src/lib/extended/style/style.ts
```diff
@@ -50,14 +51,13 @@ export class StyleDirective extends BaseDirective2 implements DoCheck {
...
+    const styles = this.nativeElement.getAttribute('style') || '';
+    this.fallbackStyles = this.buildStyleMap(styles);
   }
 
   protected updateWithValue(value: any) {
     const styles = this.buildStyleMap(value);
-    const defaultStyles = this.marshal.getValue(this.nativeElement, this.DIRECTIVE_KEY, '');
-    const fallback = this.buildStyleMap(defaultStyles);
-    this.ngStyleInstance.ngStyle = {...fallback, ...styles};
+    this.ngStyleInstance.ngStyle = {...this.fallbackStyles, ...styles};
```
ThomasBurleson: 👍

PR #940, opened 2018-12-19
feat(core): add static scss mixin

### LINKED ISSUE #783: Media queries via (s)CSS
# Feature Request

#### What is the desired behavior?
Use media query via mixin in CSS

#### What is the use-case or motivation for the desired behavior?
From what i've read in the DOC if i want a style eg a button with different background color based on breakpoints i can only do something like

`<button [ngClass.xs]="xs-button" [ngClass.xs]="md-button">HI!</button>`

and in CSS
```css
.xs-button {
  background-color: red;
}
.md-button {
  background-color: red;
}
```

This surely works but leads a lot of pollution in the markup, could you also export a mixins to do the below?

With CSS media query:

`<button>HI!</button>`

```css

button {
  background-color: yellow;
  @include media-breakpoint(xs) {
     background-color: red;
   }
  @include media-breakpoint(md) {
    background-color: blue;
   }
}
```

I can of course crate a set of mixins to do the above but would be great if this would exposed by flex-layout matching configured breakpoints..

Thanks

CaerusKaru:
Fixes #783


tools/gulp/tasks/build-release.ts
```diff
@@ -1,6 +1,29 @@
...
+
+// Matches all SCSS files in the different packages. Note that this glob is not used to build
+// the bundle. It's used to identify Sass files that shouldn't be included multiple times.
+const allScssDedupeGlob = join(buildConfig.packagesDir, '**/*.scss');
+
+// The entry-point for the scss theming bundle.
+const themingEntryPointPath = join(sourceDir, 'core', 'sass', '_layout-bp.scss');
+
+// Output path for the scss theming bundle.
+const themingBundlePath = join(releasePath, '_mq.scss');
```
ThomasBurleson: I recently started disliking the `_` prefix to indicate private. 
Perhaps we should not use such conventions for our file names?

CaerusKaru: Discussed offline that this is an unfortunate SASS convention

src/lib/core/sass/_layout-bp.scss
```diff
@@ -0,0 +1,76 @@
...
+// becomes
+//
+// @media (min-width: 0px) and (max-width: 599px) {
+//   background-color: red;
+// }
+@mixin layout-bp($bp) {
+  @if map-has-key($breakpoints, $bp) {
+    $min: map-get(map-get($breakpoints, $bp), begin);
+    $max: map-get(map-get($breakpoints, $bp), end);
+    @media (min-width: $min) and (max-width: $max) { @content; }
```
ThomasBurleson: VERY cool @CaerusKaru

CaerusKaru: :+1:

Issue #648, opened 2018-03-08
Overlapping breakpoints with ngClass doesn't apply on initial load, only works correctly on window resize


adamdport:
#### Bug, feature request, or proposal:
Bug

#### What is the expected behavior?
if I have overlapping breakpoints on a div, the nearest breakpoint is applied on initial load.

#### What is the current behavior?
The wrong breakpoint is applied on initial load, but is correct after resizing the window to another breakpoint and returning

#### What are the steps to reproduce?
https://stackblitz.com/edit/angular-flex-layout-seed-i45rb6

#### What is the use-case or motivation for changing an existing behavior?
The class that applies on load should be the correct class

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
"@angular/core": "5.2.7"
"@angular/flex-layout": "5.0.0-beta.13"
"@angular/material": "5.2.3"
Chrome 64

CaerusKaru:
Ok so there's good news and bad news. The good news is that you're definitely not crazy, and this is inconsistent behavior.

The bad news is that this is actually working exactly as its designed to work. Here's the logic: when there are competing breakpoints that are activated, the one with the widest range wins (this is documented on the Wiki, and means that lt-md > lt-sm, therefore blue). So on startup, all of the breakpoints become active at the same time and compete, and lt-md wins. BUT, when the window is resized, lt-sm is first deactivated, and then re-activated, meaning that the reactivation event triggered the green layout to appear.

Is any of this _correct_? Honestly, I have a hard time knowing. I am _very_ open to discussion on this, because it is a glaring inconsistency with how we treat breakpoints.

CaerusKaru:
We're going to target this for the next beta round, which should land in a couple weeks.

Issue #426, opened 2017-09-20
ObservableMedia: does not emit initial value in AOT mode


denver-HJS:
I've noticed a subtle difference in behavior in the ObservableMedia service when building with the CLI's AOT mode being on vs off.

I have a component property that dictates the responsive display of a section of content, and it's initialized via the subscription to the media change observable:

```js
  constructor(private mediaQueryService: ObservableMedia) { }

  ngOnInit() {
      this.initializeFormOptions();
      this.initializeWatchers();
  }

  private initializeWatchers(): void {
    this.mediaWatcher = this.mediaQueryService
      .subscribe(mediaChange => {
        this.activeMediaViewport = mediaChange.mqAlias;
        this.lineEntryDisplayMode = this.calculateComponentDisplayMode(this.activeMediaViewport);
      });
  }
```
This works fine in non-AOT mode as the active mqAlias is available when the application loads. However, in AOT mode no value is emitted until a window resize event is triggered.

Since there is no direct query to the service available to get the current mqAlias, I am left with a less desirable brute force approach of querying individual possibilities using the `isActive` API method.

alexfung888:
are you sure? Even without AOT, I don't always get an initial value of mqAlias. Not every time. I guess it is a race condition somewhere. Yes I ended up isActive()ing every breakpoint.

denver-HJS:
Hmmm, actually I haven't had trouble with the initial value outside of AOT mode, but I did notice that it wasn't emitting a value after changing router states and returning back. That being the case, I ended up using `isActive` on every breakpoint as well.

Maybe a `getActiveAlias` API method that's limited to the 'xs | sm | md | lg | xl' aliases would be a useful addition? It looks like @ThomasBurleson has added this as a milestone for the upcoming beta version though :+1:

ThomasBurleson:
@denver-HJS, @alexeagle - can you confirm these issues are STILL present with the nightly build ?

> [Installing nightly builds with npm](https://github.com/angular/flex-layout/wiki/Fast-Starts#npm-install)

denver-HJS:
I'd like to update and check, but unfortunately I'm at an impasse.

I can't seem to update anything via npm at this point (even after rolling back to v5.3.0): npm/npm/issues/18632

denver-HJS:
I'm not seeing any change using build *@angular/flex-layout@2.0.0-beta.9-0da1f88*

With AOT:

[...]

PR #946, opened 2018-12-20
fix(core): improve use of breakpoint priorities

ThomasBurleson:
Use breakpoint priority as the only sorting/scanning mechanism;
used to ensure correct MediaChange event notifications.

* prioritize breakpoints: non-overlaps hightest, lt-<xxx> lowest
* consistently sort breakpoints ascending by priority
  * highest priority === smallest range
  * remove hackery with reverse(), etc.
* memoize BreakPointRegistry findBy<xxx> lookups
* fix MatchMedia::observe() to support lazy breakpoint registration
* fix fragile logic in MediaMarshaller
  * fix breakpoint registration usage
  * clarify update/clear builder function callbacks
* fix MediaObserver breakpoint registration usage

### DIFF
--- a/src/lib/core/breakpoints/break-point-registry.ts
+++ b/src/lib/core/breakpoints/break-point-registry.ts
@@ -9,7 +9,9 @@ import {Injectable, Inject} from '@angular/core';
 
 import {BreakPoint} from './break-point';
 import {BREAKPOINTS} from './break-points-token';
+import {sortAscendingPriority} from './breakpoint-tools';
 
+type OptionalBreakPoint = BreakPoint | null;
 
 /**
  * Registry of 1..n MediaQuery breakpoint ranges
@@ -18,40 +20,21 @@ import {BREAKPOINTS} from './break-points-token';
  */
 @Injectable({providedIn: 'root'})
 export class BreakPointRegistry {
+  readonly items: BreakPoint[];
 
-  constructor(@Inject(BREAKPOINTS) private _registry: BreakPoint[]) {
-  }
-
-  /**
-   * Accessor to raw list
-   */
-  get items(): BreakPoint[] {
-    return [...this._registry];
-  }
-
-  /**
-   * Accessor to sorted list used for registration with matchMedia API
-   *
-   * NOTE: During breakpoint registration, we want to register the overlaps FIRST
-   *       so the non-overlaps will trigger the MatchMedia:BehaviorSubject last!
-   *       And the largest, non-overlap, matching breakpoint should be the lastReplay value
-   */
-  get sortedItems(): BreakPoint[] {
-    let overlaps = this._registry.filter(it => it.overlapping === true);
-    let nonOverlaps = this._registry.filter(it => it.overlapping !== true);
-
-    return [...overlaps, ...nonOverlaps];
+  constructor(@Inject(BREAKPOINTS) list: BreakPoint[]) {
+    this.items = [...list].sort(sortAscendingPriority);
   }
 
   /**
    * Search breakpoints by alias (e.g. gt-xs)
    */
-  findByAlias(alias: string): BreakPoint | null {
-    return this._registry.find(bp => bp.alias == alias) || null;
+  findByAlias(alias: string): OptionalBreakPoint {
+    return this.findWithPredicate(alias, (bp) => bp.alias == alias);
   }
 
-  findByQuery(query: string): BreakPoint | null {
-    return this._registry.find(bp => bp.mediaQuery == query) || null;
+  findByQuery(query: string): OptionalBreakPoint {
+    return this.findWithPredicate(query, (bp) => bp.mediaQuery == query);
   }
 
   /**
@@ -59,14 +42,14 @@ export class BreakPointRegistry {
    * e.g. gt-sm overlaps md, lg, and xl
    */
   get overlappings(): BreakPoint[] {
-    return this._registry.filter(it => it.overlapping == true);
+    return this.items.filter(it => it.overlapping == true);
   }
 
   /**
    * Get list of all registered (non-empty) breakpoint aliases
    */
   get aliases(): string[] {
-    return this._registry.map(it => it.alias);
+    return this.items.map(it => it.alias);
   }
 
   /**
@@ -75,6 +58,26 @@ export class BreakPointRegistry {
    * for property layoutGtSM.
    */
   get suffixes(): string[] {
-    return this._registry.map(it => !!it.suffix ? it.suffix : '');
+    return this.items.map(it => !!it.suffix ? it.suffix : '');
   }
+
+  /**
+   * Memoized lookup using custom predicate function
+   */
+  private findWithPredicate(key: string,
+      searchFn: (bp: BreakPoint) => boolean): OptionalBreakPoint {
+    let response = this.findByMap.get(key);
+    if (!response) {
+      response = this.items.find(searchFn) || null;
+      this.findByMap.set(key, response);
+    }
+    return response || null;
+
+  }
+
+  /**
+   * Memoized BreakPoint Lookups
+   */
+  private readonly findByMap = new Map<String, OptionalBreakPoint>();
 }
+
--- a/src/lib/core/breakpoints/breakpoint-tools.ts
+++ b/src/lib/core/breakpoints/breakpoint-tools.ts
@@ -65,8 +65,14 @@ export function mergeByAlias(defaults: BreakPoint[], custom: BreakPoint[] = []):
 }
 
 /** HOF to sort the breakpoints by priority */
-export function prioritySort(a: BreakPoint, b: BreakPoint): number {
+export function sortDescendingPriority(a: BreakPoint, b: BreakPoint): number {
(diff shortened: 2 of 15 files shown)


ThomasBurleson:
This includes the PR to remove deprecated APIs; to ensure optimizations work as desired.

.firebase/hosting.ZGlzdC9wYWNrYWdlcy9kZW1vLWFwcA.cache
```diff
@@ -0,0 +1,13 @@
+5.60874426946876ee868e.js,1545260628514,9734b215281b7a6e7fbc463a65b17e3dac2ee83f8f139fa7df02eabe493d21a7
```
CaerusKaru: remove, and add to gitignore?

src/lib/core/breakpoints/break-point-registry.ts
```diff
@@ -75,6 +58,25 @@ export class BreakPointRegistry {
...
   get suffixes(): string[] {
-    return this._registry.map(it => !!it.suffix ? it.suffix : '');
+    return this.items.map(it => !!it.suffix ? it.suffix : '');
   }
+
+  /**
+   * Memoized lookup using custom predicate function
+   */
+  private findWithPredicate(key: string, searchFn: (bp: BreakPoint) => boolean) {
+    let response: OptionalBreakPoint = this.findByMap.get(key) || null;
```
CaerusKaru: You don't need `|| null`, `Map.get` returns `undefined`, which will fail the check below. And then you can do `return response || null` if you care about the return type.

src/lib/core/breakpoints/break-point-registry.ts
```diff
@@ -75,6 +58,25 @@ export class BreakPointRegistry {
...
    */
   get suffixes(): string[] {
-    return this._registry.map(it => !!it.suffix ? it.suffix : '');
+    return this.items.map(it => !!it.suffix ? it.suffix : '');
   }
+
+  /**
+   * Memoized lookup using custom predicate function
+   */
+  private findWithPredicate(key: string, searchFn: (bp: BreakPoint) => boolean) {
```
CaerusKaru: Add return type

src/lib/core/breakpoints/breakpoint-tools.ts
```diff
@@ -65,8 +65,14 @@ export function mergeByAlias(defaults: BreakPoint[], custom: BreakPoint[] = []):
...
 
 /** HOF to sort the breakpoints by priority */
-export function prioritySort(a: BreakPoint, b: BreakPoint): number {
+export function sortDescendingPriority(a: BreakPoint, b: BreakPoint): number {
   const priorityA = a.priority || 0;
   const priorityB = b.priority || 0;
   return priorityB - priorityA;
 }
+
+export function sortAscendingPriority(a: BreakPoint, b: BreakPoint) {
```
CaerusKaru: Add return type

src/lib/server/server-provider.ts
```diff
@@ -165,3 +161,9 @@ function getClassName(element: HTMLElement, classMap: Map<HTMLElement, string>)
 
   return className;
 }
+
+export function sortAscendingPriority(a: BreakPoint, b: BreakPoint) {
```
CaerusKaru: Import from core?

src/lib/core/media-observer/media-observer.ts
```diff
@@ -85,9 +84,9 @@ export class MediaObserver {
    * This is needed so subscribers can be auto-notified of all standard, registered
    * mediaQuery activations
    */
-  private _registerBreakPoints() {
-    const queries = this.breakpoints.sortedItems.map(bp => bp.mediaQuery);
-    this.mediaWatcher.registerQuery(queries);
+  private watchActivations() {
```
CaerusKaru: Add return type

src/lib/core/media-marshaller/media-marshaller.ts
```diff
@@ -205,9 +203,9 @@ export class MediaMarshaller {
   updateElement(element: HTMLElement, key: string, value: any): void {
     const builders = this.builderMap.get(element);
     if (builders) {
-      const builder: Builder | undefined = builders.get(key);
-      if (builder) {
-        builder(value);
+      const updateFn: UpdateCallback = builders.get(key) as UpdateCallback;
+      if (!!updateFn) {
```
CaerusKaru: `if (updateFn !== undefined)`

src/lib/core/media-marshaller/media-marshaller.ts
```diff
@@ -188,9 +186,9 @@ export class MediaMarshaller {
   clearElement(element: HTMLElement, key: string): void {
     const builders = this.clearBuilderMap.get(element);
     if (builders) {
-      const builder: Builder | undefined = builders.get(key);
-      if (builder) {
-        builder();
+      const clearFn: ClearCallback = builders.get(key) as ClearCallback;
+      if (!!clearFn) {
```
CaerusKaru: `if(clearFn !== undefined)`

src/lib/core/match-media/match-media.ts
```diff
@@ -99,7 +118,7 @@ export class MatchMedia {
  * Private global registry for all dynamically-created, injected style tags
  * @see prepare(query)
  */
-const ALL_STYLES: {[key: string]: any} = {};
+const ALL_STYLES: { [key: string]: any } = {};
```
CaerusKaru: No spaces (per style guide)

src/lib/core/match-media/match-media.ts
```diff
@@ -38,59 +38,78 @@ export class MatchMedia {
...
+    });
+
+    return mqList ? merge(registration$, matchMedia$) : this._observable$;
   }
 
   /**
    * Based on the BreakPointRegistry provider, register internal listeners for each unique
    * mediaQuery. Each listener emits specific MediaChange data to observers
    */
   registerQuery(mediaQuery: string | string[]) {
```
CaerusKaru: Add return type

src/lib/core/match-media/match-media.ts
```diff
@@ -38,59 +38,78 @@ export class MatchMedia {
...
-    const list = Array.isArray(mediaQuery) ? Array.from(new Set(mediaQuery)) : [mediaQuery];
+    const list = Array.isArray(mediaQuery) ? mediaQuery as Array<string> : [mediaQuery];
```
CaerusKaru: Do you need the `as` here?

src/lib/core/match-media/match-media.ts
```diff
@@ -38,59 +38,78 @@ export class MatchMedia {
...
-
-    return this._observable$.pipe(
-      filter(change => (mediaQuery ? (change.mediaQuery === mediaQuery) : true))
+  observe(mqList?: string[]): Observable<MediaChange> {
+    const matchMedia$ = this._observable$.pipe(
+        filter((change: MediaChange) => {
+          return !!(mqList || []).find((query: string) => change.mediaQuery === query);
+        })
     );
+    const registration$: Observable<MediaChange> = Observable.create((observer: Observer<MediaChange>) => {  // tslint:disable-line:max-line-length
```
CaerusKaru: `Observable.create` is the anti-pattern here. Use `new Observable((observer) => {});` instead

src/lib/core/match-media/match-media.ts
```diff
@@ -38,59 +38,78 @@ export class MatchMedia {
...
-    if (mediaQuery) {
-      this.registerQuery(mediaQuery);
-    }
-
-    return this._observable$.pipe(
-      filter(change => (mediaQuery ? (change.mediaQuery === mediaQuery) : true))
+  observe(mqList?: string[]): Observable<MediaChange> {
+    const matchMedia$ = this._observable$.pipe(
+        filter((change: MediaChange) => {
+          return !!(mqList || []).find((query: string) => change.mediaQuery === query);
```
CaerusKaru: We should use `Array.includes` here. And if we can't, we should figure out why our TS config won't let us.

src/lib/core/match-media/match-media.ts
```diff
@@ -38,59 +38,78 @@ export class MatchMedia {
...
+  observe(mqList: string[]): Observable<MediaChange>;
+
   /**
    * External observers can watch for all (or a specific) mql changes.
    * Typically used by the MediaQueryAdaptor; optionally available to components
    * who wish to use the MediaMonitor as mediaMonitor$ observable service.
    *
-   * NOTE: if a mediaQuery is not specified, then ALL mediaQuery activations will
-   *       be announced.
+   * Use lazy registration...
```
CaerusKaru: Expand on this

.firebase/hosting.ZGlzdC9wYWNrYWdlcy9kZW1vLWFwcA.cache
ThomasBurleson: Done

CaerusKaru:
This is also supposed to fix #648 and #426, please add an explicit test case for the former (the latter may be trickier).

PR #948, opened 2018-12-20
fix(show-hide): account for fxLayout on same element

### LINKED ISSUE #945: FxLayout element with fxHide.xs doesn't hide on media changes
# Bug Report

`<div fxLayout fxHide.xs></div>` respects media changes and hides the element at extra-small screen sizes.

`<div fxLayout [fxHide.xs]="booleanVariable"></div>` respects media changes and hides the element at extra-small screen sizes if the `booleanVariable` is true.  This case is specific to my own code, and I want to make sure it works.

`<div fxLayout [fxHide.xs]="booleanVariable"></div>`

- [X] On initial load at extra-small screen sizes hides the element 
- [X] On resize to a larger size, displays the element
- [ ] On resize back to extra-small, **does not** hide the element.

**Note**: When the element doesn't have `fxLayout` (e.g. `<div [fxHide.xs]="booleanVariable"></div>`), it works properly.

https://stackblitz.com/edit/angular-flex-layout-seed-k1oskk?file=app%2Ftest.component.ts

It's broken.

Angular 7.1.4
Material 7.2.0
TypeScript 3.1.6

CaerusKaru:
Fixes #945

### DIFF
--- a/src/lib/extended/show-hide/hide.spec.ts
+++ b/src/lib/extended/show-hide/hide.spec.ts
@@ -10,7 +10,6 @@ import {CommonModule} from '@angular/common';
 import {ComponentFixture, TestBed, inject} from '@angular/core/testing';
 import {
   MatchMedia,
-  CoreModule,
   MockMatchMedia,
   MockMatchMediaProvider,
   MediaObserver,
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@CaerusKaru - problem still exists for:
```html
<div fxLayout fxLayoutAlign="start center" fxHide.xs></div>
```

ThomasBurleson:
@CaerusKaru - lgtm.

PR #949, opened 2018-12-21
fix(core): correctly handle lack of fallback values

### LINKED ISSUE #947: Changes not applied on window resize
# Bug Report

When I resize the window, I expect that the fxLayout.lt-md, ngStyle.gt-xs, fxFlexOrder, etc directives are applied. 

Page resize does not trigger changes in layout based on using the responsive directives.

Everything was working in beta 19, responsive layouts broken by 20, and as of 22, the layout works on page load but when resizing the page, the directives aren't applied.

In this stackblitz: https://stackblitz.com/edit/angular-flex-layout-seed-xoetvd?file=app/test.component.ts try resizing the window. The initial resize works but if you shrink it back, the style doesn't return to how it started out.

CaerusKaru:
Fixes #947

### DIFF
--- a/src/lib/flex/layout-align/layout-align.spec.ts
+++ b/src/lib/flex/layout-align/layout-align.spec.ts
@@ -363,6 +363,35 @@ describe('layout-align directive', () => {
       }, styler);
     });
 
+    it('should not fallback to value not registered with fxLayout', () => {
+      createTestComponent(`
+         <section fxLayout="row"
+                  fxLayoutAlign.lt-md="center center">
+         </section>
+      `);
+
+      matchMedia.activate('lt-md');
+      expectNativeEl(fixture).toHaveStyle({
+        'justify-content': 'center',
+        'align-items': 'center',
+        'align-content': 'center'
+      }, styler);
+
+      matchMedia.activate('md');
+      expectNativeEl(fixture).not.toHaveStyle({
+        'justify-content': 'center',
+        'align-items': 'center',
+        'align-content': 'center'
+      }, styler);
+
+      matchMedia.activate('lt-md');
+      expectNativeEl(fixture).toHaveStyle({
+        'justify-content': 'center',
+        'align-items': 'center',
+        'align-content': 'center'
+      }, styler);
+    });
+
     it('should fallback to closest overlapping value when the active mediaQuery change is not configured', () => { // tslint:disable-line:max-line-length
       createTestComponent(`
           <div  fxLayout


src/lib/extended/style/style.ts
```diff
@@ -55,12 +55,19 @@ export class StyleDirective extends BaseDirective2 implements DoCheck {
...
+  /** Add generated styles */
   protected updateWithValue(value: any) {
     const styles = this.buildStyleMap(value);
     this.ngStyleInstance.ngStyle = {...this.fallbackStyles, ...styles};
     this.ngStyleInstance.ngDoCheck();
   }
 
+  /** Remove generated styles */
+  protected clearStyles() {
+    this.ngStyleInstance.ngStyle = this.fallbackStyles;
```
willyboy: I think this will fix a bug we saw with ngStyle not updating on window resize that I hadn't gotten a chance to create a report for :)

src/lib/flex/flex/flex.ts
```diff
@@ -274,8 +274,11 @@ export class FlexDirective extends BaseDirective2 {
 
   /** Trigger a style reflow, usually based on a shrink/grow input event */
   protected triggerReflow() {
-    const parts = validateBasis(this.activatedValue, this.flexGrow, this.flexShrink);
-    this.marshal.updateElement(this.nativeElement, this.DIRECTIVE_KEY, parts.join(' '));
+    const activatedValue = this.activatedValue;
+    if (activatedValue) {
+      const parts = validateBasis(activatedValue, this.flexGrow, this.flexShrink);
```
CaerusKaru: Note to self, this should be a check for undefined, because the empty string is valid here

src/lib/core/base/base2.ts
```diff
@@ -108,7 +108,9 @@ export abstract class BaseDirective2 implements OnChanges, OnDestroy {
   /** Force trigger style updates on DOM element */
   protected triggerUpdate() {
     const val = this.marshal.getValue(this.nativeElement, this.DIRECTIVE_KEY);
-    this.marshal.updateElement(this.nativeElement, this.DIRECTIVE_KEY, val);
+    if (val !== undefined) {
+      this.marshal.updateElement(this.nativeElement, this.DIRECTIVE_KEY, val);
```
ThomasBurleson: I think we should move the checking `val !== undefined`  inside the `updateElement()` method.

src/lib/core/media-marshaller/media-marshaller.ts
```diff
@@ -286,7 +288,8 @@ export class MediaMarshaller {
         }
       }
     }
-    return bpMap.get('');
+    const lastHope = bpMap.get('');
+    return (key === undefined || lastHope && lastHope.has(key)) ? lastHope : undefined;
```
ThomasBurleson: What is this doing.. (in English).

PR #963, opened 2018-12-31
refactor(core): deprecate MediaObserver media$ to use asObservable()

ThomasBurleson:
MediaObserver is a service that provides APIs and also publishes an observable to mediaQuery activations.

* Deprecated MediaObserver `media$` property to use instead the `asObservable()`


ThomasBurleson:
Closed. Changes will be included in another PR coming soon.

PR #964, opened 2019-01-01
build: add Bazel rules

CaerusKaru:
* Add Bazel build rules for all entrypoints
* Add Bazel rules to serve the demo app
* Add goldens api guard
* Move from SCSS to CSS for most demo app components
  due to conflict with CLI and Bazel style generation
* Move testing utilities into core (though not exported with the bundle)
* Move other utilities into proper packages
* Create intermediary `BaseFlexFlowDirective` for use by `flex` and `flex-offset`
  and remove unneeded method from `BaseDirective2`


CaerusKaru:
Technically blocked on getting SSR tests to work with Bazel, but not actually blocked since we have regular SSR tests that run fine at the moment.

.circleci/config.yml
```diff
@@ -47,20 +47,39 @@ var_6: &save_cache
...
+# is used to ensure that not all upstream branches will be published as Github builds
+# (e.g. revert branches, feature branches)
+var_11: &publish_branches_filter
+  branches:
+    only:
+    - master
+    # 6.0.x, 7.1.x, etc.
+    - /\d+\.\d+\.x/
+    # 6.x, 7.x, 8.x etc
+    - /\d+\.x/
```
devversion: Not needed as there are no publish branches here. Only `master` should be fine.

src/apps/demo-app/e2e/BUILD.bazel
```diff
@@ -0,0 +1,45 @@
...
+#    tags = ["manual"],
+#)
+
+# Default target in this package is to run the e2e tests on the devserver.
+# This is a faster round-trip but doesn't exercise production optimizations like
+# code-splitting and lazy loading.
+#alias(
+#    name = "e2e",
+#    actual = "devserver_test",
+#)
```
devversion: commented stuff should be removed. Even though if it may be re-added in future PRs

src/apps/demo-app/src/BUILD.bazel
```diff
@@ -0,0 +1,140 @@
+package(default_visibility = ["//visibility:public"])
+
+load("//tools:defaults.bzl", "ng_module", "ts_library", "ts_web_test_suite")
+load("@build_bazel_rules_nodejs//:defs.bzl", "rollup_bundle", "history_server")
+load("@build_bazel_rules_nodejs//internal/web_package:web_package.bzl", "web_package")
+load("@build_bazel_rules_typescript//:defs.bzl", "ts_devserver")
+# load("//tools:sass_generate_binaries.bzl", "sass_generate_binaries")
```
devversion: Remove commented lines below and here.

src/apps/demo-app/src/app/media-query-status/BUILD.bazel
```diff
@@ -0,0 +1,50 @@
...
+        "@angular//packages/platform-browser",
+        "@angular//packages/platform-browser-dynamic/testing",
+        "@angular//packages/platform-browser/animations",
+    ],
+)
+
+ts_web_test_suite(
+    name = "test",
+    srcs = [],
+    # do not sort
```
devversion: Explain why this cannot be sorted. Similarly in other files.

Otherwise you should consider using a macro for this if it's used too often across multiple packages.

tools/gulp/tasks/aot.ts
```diff
@@ -27,7 +27,7 @@ task('aot:pre', sequenceTask(
 );
 
 task('aot:deps', [], execTask(
-  'npm', ['install'], {cwd: demoAppSource}));
+  'yarn', ['--cwd', demoAppSource]));
```
devversion: What's the issue with the `cwd` option?

tools/sass_generate_binaries.bzl
```diff
@@ -0,0 +1,17 @@
+load("@io_bazel_rules_sass//:defs.bzl", "sass_binary")
+
+# Generates multiple sass binaries based on a specified list of source files.
+# All generated sass binaries will be exposed as a filegroup that has all the
+# CSS outputs from the specified source files.
+def sass_generate_binaries(filegroup_name, source_files, sass_deps = []):
```
devversion: This is unused

.circleci/config.yml
CaerusKaru: We're going to add a 6.x branch in the future pending @ThomasBurleson

tools/gulp/tasks/aot.ts
CaerusKaru: Nothing! This resulted from a mishap where I had accidentally deleted the package.json in the app directories, and this was one attempt to fix it. But in general, I just think this is cleaner. Do you think I should change it back, or is this more a "why do this?" comment?

tools/sass_generate_binaries.bzl
CaerusKaru: It's unused now, but really needs to be used at some point in the demo app (will discuss this offline).

src/apps/demo-app/src/app/media-query-status/BUILD.bazel
CaerusKaru: Moved into the defaults.bzl macro. Great suggestion!

tools/gulp/tasks/aot.ts
devversion: Both seems fine, it's just personally I prefer to keep the `args` array limited to the desired action.

Spawning a process with the proper CWD makes this possible and works perfectly.

tools/sass_generate_binaries.bzl
devversion: "at some point" is the thing. Why not just add it then? This makes it hard to review, adds unnecessary bloat, and causes loss of context for the actual PR that uses this.

CaerusKaru: Alright, but this is definitely going to need to be addressed with some immediacy.

tools/gulp/tasks/aot.ts
CaerusKaru: Agreed, will change back

CaerusKaru:
@alexeagle @gregmagolan Can you PTAL? There's an issue with ts_devserver not inlining the static 
CSS file generated from SCSS. Would be happy to discuss over Slack if you need more information.

CaerusKaru:
SSR issues fixed thanks to @gregmagolan. Only remaining issues are with the demo-app, which might need to be addressed in a follow-up.

CaerusKaru:
Blocked on https://github.com/angular/angular/pull/28053

CaerusKaru:
Now blocked on https://github.com/angular/material2/pull/14787

CaerusKaru:
Closing this since it would likely need another rewrite to use Bazel v3.0 anyway.

PR #982, opened 2019-01-13
docs: make splitter work in IE

### LINKED ISSUE #526: (demo): FlexLayout Splitter demo not working on IE11
FlexBox Splitter not working on IE 11. 

use

#### Usage
```javascript
<div fxLayout="row" ngxSplit="row">
  <div fxFlex="30%" ngxSplitArea>
    Left
  </div>
  <div fxFlex="5px">
    <a href="#" ngxSplitHandle>...</a>
  </div>
  <div fxFlex="70%" ngxSplitArea>
    Right
  </div>
</div>
```

> See https://github.com/angular/flex-layout/issues/266 is not working.

Please provide the steps to resolve the issue.

The main reason for this issue is when we start dragging, after a moment when `SplitAreaDirective.activatedValue` starts returning values which are actually decimal numbers, in IE the value returned is `NaN`.

So, how to resolve the issue?

CaerusKaru:
Fixes #526

### DIFF
--- a/src/apps/demo-app/src/app/github-issues/split/split-handle.directive.ts
+++ b/src/apps/demo-app/src/app/github-issues/split/split-handle.directive.ts
@@ -23,7 +23,10 @@ export class SplitHandleDirective {
     };
 
     const mousedown$ = fromEvent(ref.nativeElement, 'mousedown').pipe(map(scanPositionDelta));
-    const mousemove$ = fromEvent(window, 'mousemove').pipe(throttleTime(50), map(scanPositionDelta));
+    const mousemove$ = fromEvent(window, 'mousemove').pipe(
+      throttleTime(50),
+      map(scanPositionDelta)
+    );
     const mouseup$ = fromEvent(window, 'mouseup');
 
     this.drag = mousedown$.pipe(switchMap(() => mousemove$.pipe(takeUntil(mouseup$))));
--- a/src/apps/demo-app/src/app/github-issues/split/split-handle.directive.ts
+++ b/src/apps/demo-app/src/app/github-issues/split/split-handle.directive.ts
@@ -1,7 +1,7 @@
 import {Directive, ElementRef, Inject, Output} from '@angular/core';
 import {DOCUMENT} from '@angular/common';
 import {fromEvent, Observable} from 'rxjs';
-import {map, switchMap, takeUntil, tap} from 'rxjs/operators';
+import {map, switchMap, takeUntil, throttleTime} from 'rxjs/operators';
 
 @Directive({
   selector: '[ngxSplitHandle]',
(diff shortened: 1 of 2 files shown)


src/apps/demo-app/src/app/github-issues/split/split.directive.ts
```diff
@@ -62,24 +60,17 @@ export class SplitDirective implements AfterContentInit, OnDestroy {
...
-  valueToPixel(value: string | number, parentWidth: number): number {
-    const isPercent = () => String(value).indexOf('px') < 0;
-    let size = parseFloat(String(value));
-    if (isPercent()) {
-      size = parentWidth * (size / 100);  // Convert percentage to actual pixel float value
-    }
-    return size;
-  }
+/** Convert the pixel or percentage value to a raw pixel float value */
+function valueToPixel(value: string, parentWidth: number): number {
```
ThomasBurleson: This does not work. Consider `valueToPixel('327')` which returns **NaN**.
You need
```ts
function valueToPixel(value: string, parentWidth = 100): number {
   ...
}

Comment (CaerusKaru) [reply src/apps/demo-app/src/app/github-issues/split/split.directive.ts]: `parentWidth` is not optional though?

Comment (ThomasBurleson) [reply src/apps/demo-app/src/app/github-issues/split/split.directive.ts]: True that... I still contend my version is more clear. It is a number and defaults to `100`.

File: src/apps/demo-app/src/app/github-issues/split/split-handle.directive.ts [hunk 2]
```diff
@@ -14,11 +14,24 @@ export class SplitHandleDirective {
   @Output() drag: Observable<{ x: number, y: number }>;
 
   constructor(ref: ElementRef, @Inject(DOCUMENT) _document: any) {
-    const getMouseEventPosition = (event: MouseEvent) => ({x: event.movementX, y: event.movementY});
+    let prevX = 0;
```
ThomasBurleson: I suggest this is a improved version:

```ts
  let last = {x: 0, y: 0};
  const getPositionDelta = (event: MouseEvent) => { 
    const current = {x: event.screenX, y: event.screenY};   
    const delta = { x: current.x - last.x, y: current.y - last.y };
    
    last = current;
    return delta;
  }
  
  const mousedown$ = fromEvent(target, 'mousedown').pipe( map(getPositionDelta) );
  const mousemove$ = fromEvent(window, 'mousemove').pipe( throttleTime(50), map(getPositionDelta) );
  const mouseup$ = fromEvent(window, 'mouseup');

  return mousedown$.pipe(switchMap(() => mousemove$.pipe( takeUntil(mouseup$) )));
```

PR #987, opened 2019-01-13
fix(sass): update stylesheet with correct keys

CaerusKaru:


fabiocarneiro:
Nice!

mackelito:
When can we expect this to be released? I'm really eager to try out the overlapping keys :)

CaerusKaru:
It’ll be included in the next beta release, currently tracking for mid-February.

YakirSmartag:
Hi, does anyone knows when the next version will be released (bate.24)? I see little activity in this repository

jarekw87:
@CaerusKaru please fix the mixin:

```
@mixin layout-bp($bp) {
  @if map-has-key($breakpoints, $bp) {
    $min: map-get(map-get($breakpoints, $bp), begin);
    $max: map-get(map-get($breakpoints, $bp), end);
    @media (min-width: $min) and (max-width: $max) { @content;  background-color: green;}
  }
  @else if map-has-key($overlapping-gt, $bp) {
    $min: map-get($overlapping-gt, $bp); // $breakpoints change to $overlapping-gt
    @media (min-width: $min) { @content; background-color: pink; }
  }
  @else if map-has-key($overlapping-lt, $bp) {
    $max: map-get($overlapping-lt, $bp); //$breakpoints change to $overlapping-lt
    @media (max-width: $max) { @content; background-color: blue;}
  }
}
```

When can we expect this to be released? :)

PR #998, opened 2019-01-15
feat(core): move MatchMedia to internal scope

CaerusKaru:
BREAKING CHANGE:

* `MatchMedia` is no longer exported as a public utility. Please use `MediaObserver` instead. 
* `ServerMatchMedia` is no longer exported at all

### DIFF
--- a/src/lib/server/server-match-media.ts
+++ b/src/lib/server/server-match-media.ts
@@ -7,9 +7,7 @@
  */
 import {DOCUMENT} from '@angular/common';
 import {Inject, Injectable, NgZone, PLATFORM_ID} from '@angular/core';
-
-import {BreakPoint} from '../breakpoints/break-point';
-import {MatchMedia} from './match-media';
+import {BreakPoint, ɵMatchMedia as MatchMedia} from '@angular/flex-layout/core';
 
 /**
  * Special server-only class to simulate a MediaQueryList and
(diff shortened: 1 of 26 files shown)


src/lib/server/server-match-media.ts
```diff
@@ -7,9 +7,7 @@
  */
 import {DOCUMENT} from '@angular/common';
 import {Inject, Injectable, NgZone, PLATFORM_ID} from '@angular/core';
-
-import {BreakPoint} from '../breakpoints/break-point';
-import {MatchMedia} from './match-media';
+import {BreakPoint, ɵMatchMedia as MatchMedia} from '@angular/flex-layout/core';
```
ThomasBurleson: Do we really want to import for `@angular/flex-layout` ?

CaerusKaru: Yes, as explained offline. This is how we handle imports throughout the library due to TS conflicts if we did everything using relative paths.

willyboy:
We were using mockMatchMedia to activate different breakpoints in tests. What's the preferred approach now? I could open a docs PR with a bit of direction.

MaT1g3R:
Hello, is there any updates on this? We are using `MockMatchMedia` too and we would like to know what is the preferred approach now as well, thanks.

Issue #1001, opened 2019-01-20
Sub-Pixel calculation issue


charsleysa:
# Bug Report

#### What is the expected behavior?
No problems caused by subpixel calculations by browsers.

#### What is the current behavior?
If the browser ends up with the width of the viewport set to something like 1279.5px then neither the md nor the lg breakpoints match.

#### What are the steps to reproduce?
Here's a codepen showing the issue using basic CSS but same applies to this library: https://codepen.io/anon/pen/VqoGXP

![image](https://user-images.githubusercontent.com/1219661/51436861-d0619000-1cf9-11e9-81ab-85e25e3199b2.png)

#### What is the use-case or motivation for changing an existing behavior?
This really starts to apply to higher density screens with scaling turned of (such as a 2160p monitor with Windows set to 150% scaling and using 2 browsers on each half of the screen).

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
All.

#### Is there anything else we should know?
The fix is pretty simple, just add `.9999` to the end of all the pixel values that are 1px less than the next break point (e.g. the md breakpoint max-width would be set to 1279.9999px).

CaerusKaru:
In an absolute ideal world, we would switch the syntax for our media queries to use Media Query Level 4 syntax (`< 1280px` and `>= 1280px`, etc). However, this is not an ideal world because that wouldn't work on IE11 and most older browsers.

I'll discuss modifying the media queries with @ThomasBurleson. In the meantime, if you'd like to submit a PR for this, we'd be happy to review it when ready.

PR #1006, opened 2019-01-23
update(core): align breakpoints with those used in CDK

### LINKED ISSUE #685: feat: switch to CDK BreakpointsObserver
We need to migrate to the CDK's layout engine. This means the following:

- [ ] Switch to using `BreakpointsObserver` in `MediaMonitor`
- [x] Deprecate `ObservableMedia`/`MediaService`
- [ ] Deprecate `MatchMedia`
- [ ] Add server wrapper for `BreakpointsObserver` and `MediaMatcher`
- [ ] Update the CDK layout engine as necessary

ThomasBurleson:
* Ensure standard breakpoints mediaQueries are aligned with those in the CDK
* Update MediaObserver
  * isActive() enhanced to support list of aliases to determine if any match
  * properly disconnects subscribers when destroyed

The `@angular/cdk` BreakpointObserver will not replace `@angular/flex-layout` MediaObserver. **MediaObserver** is an enhanced version that notifies subscribers of activations for standard AND **overlapping (lt-xxx, gt-xxx)**  breakpoints.

> Note: Developers should use MediaObserver (not use MatchMedia) service to observe breakpoint activations! MediaObserver is the recommended service to use for application developers; MatchMedia should be considered a private service.

Fixes #685. Refs #1001.

### DIFF
--- a/src/lib/core/breakpoints/data/break-points.ts
+++ b/src/lib/core/breakpoints/data/break-points.ts
@@ -13,52 +13,52 @@ import {BreakPoint} from '../break-point';
 export const DEFAULT_BREAKPOINTS: BreakPoint[] = [
   {
     alias: 'xs',
-    mediaQuery: 'screen and (min-width: 0px) and (max-width: 599.9999px)',
+    mediaQuery: 'screen and (min-width: 0px) and (max-width: 599.99px)',
     priority: 1000,
   },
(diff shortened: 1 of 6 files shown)


ThomasBurleson:
@CaerusKaru - ping for review

PR #1087, opened 2019-06-24
fix(server): breakpoints can be used by MediaObserver

Chris3773:
MediaObserver on the server was not observing the breakpoints at the same lifecycle as the browser. This change moves the active breakpoints check from `BEFORE_APP_SERIALIZED` to the `ServerMatchMedia` service.

### DIFF
--- a/src/lib/server/server-match-media.ts
+++ b/src/lib/server/server-match-media.ts
@@ -7,7 +7,13 @@
  */
 import {DOCUMENT} from '@angular/common';
 import {Inject, Injectable, NgZone, PLATFORM_ID} from '@angular/core';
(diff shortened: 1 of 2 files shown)


src/lib/server/server-match-media.ts
```diff
@@ -26,7 +25,7 @@ export class ServerMediaQueryList implements MediaQueryList {
     return this._mediaQuery;
   }
 
-  constructor(private _mediaQuery: string) {}
+  constructor(private _mediaQuery: string, private _isActive = false) {}
```
CaerusKaru: Change this back, otherwise looks good :+1:

CaerusKaru:
Just needs a fix to pass the lint CI and should be good to go. Thanks for this!

Issue #1059, opened 2019-05-09
MediaObserver asObservable() emit wrong events on breakpoints


eprevot:
# Bug Report

#### What is the expected behavior?
When I resize the screen to 960px I should have a MediaChange event with mqAlias="md"

#### What is the current behavior?
When I resize the screen to 960px I have (among others) 2 MediaChange events with mqAlias="md" and mqAlias="sm"

#### What are the steps to reproduce?
A component with
```
public ngOnInit() {
    this.watcher = this.mediaObserver.asObservable()
      .subscribe((changes: MediaChange[]) => {
        console.log(changes)
      })
}
```

#### What is the use-case or motivation for changing an existing behavior?
When I resize slowly, MediaObserver.media$ does not always emit an event when breakpoints are crossed. I think it may be related.

#### Is there anything else we should know?
It may be related to the recent changes about the breakpoints (i.e. max-width: 959.99 instead of 959)
fxLayout version: 7.0.0-beta.24

PapaNappa:
Same here.

This is my code

    mediaObserver.asObservable().subscribe(changes => {
      console.log(changes.map(c => c.mqAlias));
    });

In Chrome, when I resize fast (from lg to md and back), I get the following outputs

```
(5) ["lg", "lt-xl", "gt-md", "gt-sm", "gt-xs"] # after page load
(5) ["md", "lt-lg", "lt-xl", "gt-sm", "gt-xs"] # after making window smaller
(5) ["md", "lt-lg", "lt-xl", "gt-sm", "gt-xs"] # same event at the same time
(5) ["lg", "lt-xl", "gt-md", "gt-sm", "gt-xs"] # after making window larger
(5) ["lg", "lt-xl", "gt-md", "gt-sm", "gt-xs"] # same event at the same time
```

This is expected (except that all events are emitted twice).

When resizing slowly, I get the following outputs

```
(5) ["lg", "lt-xl", "gt-md", "gt-sm", "gt-xs"] # after page load
(7) ["md", "lt-lg", "lg", "lt-xl", "gt-md", "gt-sm", "gt-xs"] # after making window smaller
(7) ["md", "lt-lg", "lg", "lt-xl", "gt-md", "gt-sm", "gt-xs"] # same event at the same time
(7) ["md", "lt-lg", "lg", "lt-xl", "gt-md", "gt-sm", "gt-xs"] # after making window larger
[...]

PR #1113, opened 2019-08-19
fix(core): correctly emit MediaChange events

paulotokimatu:
Related to #1059 

**Before:**

https://imgur.com/CjYklut

Notice that both `sm` and `md` MediaQueries are activated.

**After:**

https://imgur.com/qKH32Lg

### DIFF
--- a/src/lib/core/media-observer/media-observer.ts
+++ b/src/lib/core/media-observer/media-observer.ts
@@ -163,7 +163,7 @@ export class MediaObserver implements OnDestroy {
         .observe(this.hook.withPrintQuery(mqList))
         .pipe(
             filter((change: MediaChange) => change.matches),
-            debounceTime(0, asapScheduler),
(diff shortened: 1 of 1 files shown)


paulotokimatu:
@ThomasBurleson Do you think there is any issue reverting the dueTime from 0 to 10?

paulotokimatu:
@CaerusKaru Very sorry for bothering you, just pinging for feedback.

epelc:
@paulotokimatu are you sure this is the correct way to fix this? It looks like `debounceTime` was used to switch the scheduler. Changing 0 to 10ms seems like it would have unintended consequences ie make it racy on slower devices.

CaerusKaru:
I agree this isn't the correct solution. I appreciate your putting the time in to create this PR, and would be happy to work with you to get the correct fix in.

PR #1114, opened 2019-08-30
build: add Directive decorators and export all base classes

CaerusKaru:
Ref https://github.com/angular/angular/issues/31956


CaerusKaru:
cc @Splaktar

PR #1121, opened 2019-09-10
ci: update lock-closed to only run once per day

josephperrott:
Now that the backlog of inactive closed issues/prs has been locked, we should only run the action once per day.


mackelito:
Is there some way to mute issues that have been locked?

CaerusKaru:
If you mean issues that are about to be blocked, unfortunately no. Just for me, I’ve set up an email filter for all locking notifications (with the sender being the bot)

PR #1126, opened 2019-09-20
fix(core) Universal server renders breakpoint @media queries properly

### LINKED ISSUE #1065: Breakpoint fallback broken for CSS @media queries with Universal server rendering
# Bug Report
I'm trying to migrate to Universal SSR. Flex Layout @media query breakpoints seem to be broken. Is there a workaround for now?

Universal SSR generates a <style> block that should mimic the browser-side JS runtime behavior, but it's consistently not.

**Example A**: xs shows 1, 2, 3 in a column, sm shows 1,2 on one line and 3 on the next, and md/lg/xl shows 1, 2, 3 in a row.
```
<div fxLayout="row wrap" fxLayoutAlign="center start">
  <div fxFlex="100%" fxFlex.gt-xs="50%" fxFlex.gt-sm="33.33%"> 1 </div>
  <div fxFlex="100%" fxFlex.gt-xs="50%" fxFlex.gt-sm="33.33%"> 2 </div>
  <div fxFlex="100%" fxFlex.gt-xs="50%" fxFlex.gt-sm="33.33%"> 3 </div>
</div>
```

**Example B**: xs/sm shows 1, 2, 3 in a column, md/lg shows 1,2 on one line and 3 on the next, and xl shows 1, 2, 3 in a row.
```
<div fxLayout="row wrap" fxLayoutAlign="center start">
  <div fxFlex="100%" fxFlex.gt-sm="50%" fxFlex.gt-lg="33.33%"> 1 </div>
  <div fxFlex="100%" fxFlex.gt-sm="50%" fxFlex.gt-lg="33.33%"> 2 </div>
  <div fxFlex="100%" fxFlex.gt-sm="50%" fxFlex.gt-lg="33.33%"> 3 </div>
</div>
``` 

Breakpoints aren't properly calculated, showing the wrong layout on initial page render, which then jarringly corrects itself once JS has loaded.

**Example A**: xs shows 1, 2, 3 in a column, sm/md/lg/xl shows in a row.

**Example B**: all screen sizes shows 1, 2, 3 in a row.

It seems the @media queries are not generated correctly. Here's the CSS that Example A generates for one of the divs. The gt-xs generates a media query at the top (`(min-width: 600px)`) that is then overridden by the sm below it (`(min-width: 600px) and (max-width: 959px)`):

```
<style class="flex-layout-ssr">
@media all { .flex-layout-355 { max-width:100%; } }
@media screen and (min-width: 600px) { .flex-layout-355 { max-width:50%; } }
@media screen and (min-width: 960px) { .flex-layout-355 { max-width:33.33%; } }
@media screen and (min-width: 1280px) { .flex-layout-355 { max-width:33.33%; } }
@media screen and (min-width: 1920px) { .flex-layout-355 { max-width:33.33%; } }
@media screen and (min-width: 1920px) and (max-width: 5000px) { .flex-layout-355 { max-width:33.33%; } }
@media screen and (max-width: 1919px) { .flex-layout-355 { max-width:33.33%; } }
@media screen and (min-width: 1280px) and (max-width: 1919px) { .flex-layout-355 { max-width:33.33%; } }
@media screen and (max-width: 1279px) { .flex-layout-355 { max-width:33.33%; } }
@media screen and (min-width: 960px) and (max-width: 1279px) { .flex-layout-355 { max-width:33.33%; } }
@media screen and (max-width: 959px) { .flex-layout-355 { max-width:33.33%; } }
@media screen and (min-width: 600px) and (max-width: 959px) { .flex-layout-355 { max-width:33.33%; } }
@media screen and (max-width: 599px) { .flex-layout-355 { max-width:100%; } }
@media screen and (min-width: 0px) and (max-width: 599px) { .flex-layout-355 { max-width:100%; } }
</style>
```

Serve app via Angular Universal (FlexLayoutServerModule in app.server.module.ts). Add this to a page:

```
<div fxLayout="row wrap" fxLayoutAlign="center start">
  <div fxFlex="100%" fxFlex.gt-xs="50%" fxFlex.gt-sm="33.33%"> 1 </div>
  <div fxFlex="100%" fxFlex.gt-xs="50%" fxFlex.gt-sm="33.33%"> 2 </div>
  <div fxFlex="100%" fxFlex.gt-xs="50%" fxFlex.gt-sm="33.33%"> 3 </div>
</div>
```

Disable JS in browser. Load page.

To render page accurately. Without this, Flex Layout with Angular Universal displays layouts improperly.

FlexLayout 7.0.0-beta.24.
Angular 7.2.15.
Material 7.3.7.
macOS 10.14.3.
TypeScript 3.2.4.
Firefox 66 and Chrome 74.

bunnyfly:
Responsive breakpoints were generating broken CSS @media queries that
didn't respect breakpoint priority. This fixes the generated CSS on the
server. See issue #1065 for details.

Fixes #1065

### DIFF
--- a/src/lib/core/media-marshaller/media-marshaller.ts
+++ b/src/lib/core/media-marshaller/media-marshaller.ts
@@ -5,7 +5,8 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {Injectable} from '@angular/core';
+import {Inject, Injectable, PLATFORM_ID} from '@angular/core';
+import {isPlatformServer} from '@angular/common';
 
 import {merge, Observable, Subject, Subscription} from 'rxjs';
 import {filter, tap} from 'rxjs/operators';
@@ -58,7 +59,8 @@ export class MediaMarshaller {
 
   constructor(protected matchMedia: MatchMedia,
               protected breakpoints: BreakPointRegistry,
-              protected hook: PrintHook) {
+              protected hook: PrintHook,
+              @Inject(PLATFORM_ID) private platformId: Object) {
     this.observeActivations();
   }
 
@@ -175,7 +177,8 @@ export class MediaMarshaller {
   updateStyles(): void {
     this.elementMap.forEach((bpMap, el) => {
       const keyMap = new Set(this.elementKeyMap.get(el)!);
-      let valueMap = this.getActivatedValues(bpMap);
+      let valueMap = isPlatformServer(this.platformId) ?
+          this.getServerActivatedValues(bpMap) : this.getActivatedValues(bpMap);
 
       if (valueMap) {
         valueMap.forEach((v, k) => {
@@ -185,11 +188,12 @@ export class MediaMarshaller {
       }
 
       keyMap.forEach(k => {
-        valueMap = this.getActivatedValues(bpMap, k);
+        valueMap = isPlatformServer(this.platformId) ?
+            this.getServerActivatedValues(bpMap, k) : this.getActivatedValues(bpMap, k);
         if (valueMap) {
           const value = valueMap.get(k);
           this.updateElement(el, k, value);
-        } else {
+        } else if (!isPlatformServer(this.platformId)) {
           this.clearElement(el, k);
         }
       });
@@ -324,6 +328,22 @@ export class MediaMarshaller {
     }
     const lastHope = bpMap.get('');
     return (key === undefined || lastHope && lastHope.has(key)) ? lastHope : undefined;
+  }
+
+  /**
+   * get the bpMap's value for the current breakpoint on the server
+   * @param bpMap
+   * @param key
+   */
+  private getServerActivatedValues(bpMap: BreakpointMap, key?: string): ValueMap | undefined {
+    // Unlike on the browser, only the currently activated breakpoint should be used on the server
+    // to generate styles, otherwise all possible breakpoint styles will be present on Universal's
+    // render and break responsive layouts.
+    const valueMap = bpMap.get(this.activatedAlias);
+    if (valueMap && (key === undefined || valueMap.has(key))) {
+      return valueMap;
+    }
+    return undefined;
   }
 
   /**
--- a/src/apps/universal-app/src/app/app.component.html
+++ b/src/apps/universal-app/src/app/app.component.html
@@ -41,4 +41,26 @@
       </div>
     </div>
   </div>
+
+  <h2>Responsive layout demo 1</h2>
+  <p>
+    Disable JS to see page as rendered by Universal.
+    You should see for xs & sm: a column, md & lg: 1 and 2 in a row with 3 below, xl: a row.
+  </p>
+  <div fxLayout="row wrap" fxLayoutAlign="center start">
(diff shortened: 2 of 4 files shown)


bunnyfly:
Oops. Didn't know about the lib:test:ssr that's failing. Will work on a fix later today!

Meanwhile, I duplicated the existing tests into browser and server versions and didn't notice anything else about the testing of the changed code that would actually be tested. If the duplication is unnecessary or another path is preferred, let me know.

bunnyfly:
Hi @CaerusKaru or anyone more familiar with this codebase. Mind giving a bit of help here?

The layout-align tests are failing with this change. For example, https://github.com/angular/flex-layout/blob/master/src/lib/flex/layout-align/layout-align.spec.ts#L318 fails saying the max-width style isn't present on the element.

This makes sense, since the layout should be styled via CSS classes on the server (not inline style), and the unit tests are running in server. Ideally the unit tests would have both server and browser versions, but I can't seem to get either working.

I've tried a few things which all fail, notably:

- Use `toHaveCSS` instead of `toHaveStyle`, but it seems the CSS is empty too.
- Remove the `SERVER_TOKEN` to emulate a browser, since the behavior there shouldn't be altered. But this doesn't do anything.
- Remove the `SERVER_TOKEN` *and* provide `PLATFORM_ID` with `'browser'`. But this results in `ReferenceError: navigator is not defined`

Before I keep bashing my head against the wall, I wanted to check in and see what you (or anyone more familiar with the codebase) would recommend. Thanks!

src/lib/core/media-marshaller/media-marshaller.ts
```diff
@@ -326,6 +330,22 @@ export class MediaMarshaller {
     return (key === undefined || lastHope && lastHope.has(key)) ? lastHope : undefined;
   }
 
+  /**
+   * get the bpMap's value for the current breakpoint on the server
+   * @param bpMap
+   * @param key
+   */
+  private getServerActivatedValues(bpMap: BreakpointMap, key?: string): ValueMap | undefined {
+    // Unlike on the browser, only the currently activated breakpoint should be used on the server
```
CaerusKaru: This is what I don't understand about your PR; I'm hoping you can clarify. The _intention_ is to generate all possible breakpoint styles on the server, because you don't _know_ what breakpoint is active on the client. The server has no active breakpoints! To that end, there's no concept of a browser, browser window, etc. We use the benefit of CSS (specifically the "Cascading") to give priority to the more exclusive breakpoints. If there's an issue there, the generation algorithm is in the `server` package. But I don't see how this fixes anything.

bunnyfly: Haha. That makes two of us 😂

This is just based on the patch @wsfriday-sfd created in issues 1065. I don't get how, but universal rendering's been working on my own project ever since I used my own version with @wsfriday-sfd's patch. I was hoping just submitting it as a PR would allow me to upgrade and keep up beyond Angular 6, but it seems it might not actually be solving the underlying problem.

If @wsfriday-sfd is around, perhaps they could clarify how/why this works in practice. Otherwise, I'll try to take a stab at actually understanding and solving where the bug is. Unfortunately I don't have tons of free time, but I'll try to look into it again in the next couple weeks.

CaerusKaru:
I'm going to close this for now. If there's a more concrete solution/explanation, we can reopen.

PR #1142, opened 2019-11-06
build: fix gulp setup not working with node v12

### LINKED ISSUE #1079: Document node 11 is required to build(for gulp 3.x)
Document that you must use node <12(ie 11.x) in order to build the project since gulp 3.X is being used(gulp 4 supports node 12).

The wiki mentions you must have > 8.x but no mention that > 12.x will no longer work. https://github.com/angular/flex-layout/wiki/Developer-Setup

I suggest using `nvm` since angular 8 projects now use node 12 so you have to switch between versions if you are working on this vs an angular 8 project using this.

CaerusKaru:
See angular/angular/pull/33348

Fixes #1079

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@
 # To validate changes, use an online parser, eg.
 # http://yaml-online-parser.appspot.com/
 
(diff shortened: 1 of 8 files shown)


CaerusKaru:
FYI @devversion

PR #1143, opened 2019-11-06
build: upgrade to Angular and CDK v9

CaerusKaru:
Closes #1151


src/apps/universal-app/server.ts.bak
```diff
@@ -0,0 +1,43 @@
+// These are important and needed before anything else
```
alan-agius4: the bak file can be removed

alan-agius4:
I am not seeing any changes to the tsconfig, which should have been done when updating the CLI.

PR #1168, opened 2020-01-02
fix(show-hide): display set to "initial" value if parent has no display set

### LINKED ISSUE #1163: fxHide/fxShow incosistency with SSR and client rendered app
# Bug Report

`fxHide.x` should behave the same when rendered by SSR and client

`fxHide.x` doesn't set the `display: block` in the CSS media query, it leaves it empty and items remain hidden.

Try this code rendered by SSR with JS off in a browser and rendered by the browser to see the problem.

```
    <div class="rounded-thingies" fxHide fxShow.gt-sm>
      <!-- DESKTOP -->
    </div>
    
    <div class="rounded-thingies" fxHide fxShow.lt-md>
      <!-- MOBILE & TABLET -->
    </div>
```

Angular: `8.2.14`
FlexLayout: `8.0.0-beta.26`

vmasek:
When SSR creates media screen styles it sets the default display to `parent.display`, but if the parent has its `display` property unset, it results in an empty display which breaks the correct functionality as the next closest match is made.

Fixed by setting display to `initial` as default instead of an empty value.

fixes #1163

### DIFF
--- a/src/lib/extended/show-hide/show-hide.ts
+++ b/src/lib/extended/show-hide/show-hide.ts
@@ -37,7 +37,7 @@ export interface ShowHideParent {
 export class ShowHideStyleBuilder extends StyleBuilder {
   buildStyles(show: string, parent: ShowHideParent) {
     const shouldShow = show === 'true';
-    return {'display': shouldShow ? parent.display : 'none'};
+    return {'display': shouldShow ? parent.display || 'initial' : 'none'};
   }
 }
 


vmasek:
As I see the tests fail, I'm not sure if this actually breaks something or how to altern the tests to match this behaviour. Please leave feedback on what do you think and I'll try to fix it.

gurkan0791:
It's my understanding that I should add "fxLayout" to all "mat-header-cell" and "mat-cell" tags?Otherwise it will be corrupted due to "display: initial".
For Example, [Stackblitz](https://stackblitz.com/edit/angular-jnsl2n?file=src%2Fapp%2Ftable-basic-flex-example.html)

CaerusKaru:
@gurkan0791 I’m a little confused by your comment; can you open an issue with your StackBlitz along with what you think the expected behavior should be?

CaerusKaru:
Tracking in #1250

PR #1177, opened 2020-01-26
build: upgrade to latest Angular and CDK RC

CaerusKaru:


src/lib/flex/layout-gap/layout-gap.ts
```diff
@@ -114,14 +113,12 @@ export class LayoutGapDirective extends BaseDirective2 implements AfterContentIn
     return buffer;
   }
 
-  constructor(protected elRef: ElementRef,
+  constructor(elRef: ElementRef,
               protected zone: NgZone,
```
Splaktar: Why is `protected` OK on this line but not on the above line?

src/lib/flex/layout-gap/layout-gap.ts
```diff
@@ -114,14 +113,12 @@ export class LayoutGapDirective extends BaseDirective2 implements AfterContentIn
...
               protected directionality: Directionality,
               protected styleUtils: StyleUtils,
```
Splaktar: It seems like `protected` was removed from most of the other instances of `styleUtils`, but not this one. Can you share why? I'm just trying to better understand these changes.

src/lib/core/style-utils/style-utils.ts
```diff
@@ -16,8 +16,8 @@ import {LAYOUT_CONFIG, LayoutConfigOptions} from '../tokens/library-config';
 @Injectable({providedIn: 'root'})
 export class StyleUtils {
 
-  constructor(@Optional() private _serverStylesheet: StylesheetMap,
```
Splaktar: Can you explain why all of the instances of `@Optional()` were removed? Due to TypeScript? Ivy?

src/lib/extended/class/class.ts
```diff
@@ -24,16 +34,18 @@ export class ClassDirective extends BaseDirective2 implements DoCheck {
...
+              styler: StyleUtils,
+              marshal: MediaMarshaller,
+              iterableDiffers: IterableDiffers,
+              keyValueDiffers: KeyValueDiffers,
+              renderer2: Renderer2,
               @Optional() @Self() protected readonly ngClassInstance: NgClass) {
     super(elementRef, null!, styler, marshal);
     if (!this.ngClassInstance) {
       // Create an instance NgClass Directive instance only if `ngClass=""` has NOT been defined on
       // the same host element; since the responsive variations may be defined...
```
Splaktar: It's not clear that this comment is still relevant? I don't see the check for the empty string not being defined?

Also there appears to be an extra `instance` in the first line of the comment.

src/lib/extended/style/style.ts
```diff
@@ -44,19 +46,20 @@ export class StyleDirective extends BaseDirective2 implements DoCheck {
...
+              differs: KeyValueDiffers,
+              renderer2: Renderer2,
               @Optional() @Self() private readonly ngStyleInstance: NgStyle,
-              @Optional() @Inject(SERVER_TOKEN) serverLoaded: boolean,
+              @Inject(SERVER_TOKEN) serverLoaded: boolean,
               @Inject(PLATFORM_ID) platformId: Object) {
     super(elementRef, null!, styler, marshal);
     if (!this.ngStyleInstance) {
       // Create an instance NgStyle Directive instance only if `ngStyle=""` has NOT been
       // defined on the same host element; since the responsive variations may be defined...
```
Splaktar: Same questions here as for `ngClassInstance` above.

Issue #1161, opened 2019-12-17
Simplify gdColumns use 


romelgomez:
# Proposal

#### What is the summary of the proposal?

Setting gdColumns is verbose and only works on page reload, if the page is resized it will keep the same grid.   
 
https://github.com/angular/flex-layout/blob/ec7b57eb6adf59ecfdfff1de5ccf1ab2f6652ed3/src/lib/grid/columns/columns.ts#L79

```html
gdColumns="auto"  gdColumns.xs="auto"  gdColumns.sm="auto"  gdColumns.md="auto auto"  gdColumns.lg="auto auto auto" gdColumns.xl="auto auto auto" 
  gdColumns.lt-sm="auto"  gdColumns.lt-md="auto auto"  gdColumns.lt-lg="auto auto auto"  gdColumns.lt-xl="auto auto auto"  
  gdColumns.gt-xs="auto"  gdColumns.gt-sm="auto"  gdColumns.gt-md="auto auto"  gdColumns.gt-lg="auto auto"
```

I found a directive that works* more simple but make use of MatGridListModule 

https://stackoverflow.com/a/55407217/2513972

I figure out how to modified it to work only with flex-layout but only I make it work with a component. 

I am sharing a functional example:  https://stackblitz.com/edit/angular-nfng3n 

I find some directives can set attributes in other directives implementation, the stackoverflow example do just that, and I do like do the same thing with flex-layout to have a more reusable directive, example:  

```html
<div fxLayout="row" fxLayoutAlign="space-between none" fxLayoutGap="4px"  [appResponsiveColumns]="{xs: 'auto', sm: 'auto', md: 'auto auto', lg: 'auto auto auto', xl: 'auto auto auto'}">
```

And the appResponsiveColumns directive take care of update gdColumns attribute

CaerusKaru:
Issues with page resizing are covered in #1059, and are limited to Chrome. Since this is behavior we _do_ support, but is broken at the browser level, I'm closing this issue. You can track updates in the tracking ticket.

Issue #1141, opened 2019-11-06
[Design] Style reconciliation


CaerusKaru:
There's a tendency for directives to run into conflicts, e.g. fxShow + fxLayout. Each directive has a conflicting idea of what styles to generate for a particular node, so we need a way to reconcile these at the top-level, with priorities built-in.

CaerusKaru:
Closing as part of the RFC.

Issue #1130, opened 2019-09-25
Was fxShow and fxHide designed to work with gd?


RyanTibbetts-OnScale:
# Bug Report

#### What is the expected behavior?

`fxShow` and `fxHide` with responsive tags should hide or show elements.

#### What is the current behavior?

So I'm not certain if this was the intention, but it looks like when setting `fxShow` or `fxHide` with responsive tags that `gd` overwrites or overrides the `display: none;` set by `fxShow` and `fxHide`.

#### What are the steps to reproduce?

To replicate what I'm seeing, simply take my code below (since the Stack Blitz doesn't work right now) and put it into any existing angular project using angular flex layout. When you're at screen size `gt-sm` you'll see three blocks for each case: the blocks in a `gd` and the blocks not in a `gd`. They are roughly identical except that one is using `[fxShow.lt-md]="booleanValue"` alongside a `gd` and the other is using a `[fxShow.lt-md]="booleanValue"` without a `gd`.

If you start at `gt-sm` and resize your window to `lt-md` the one in the `gd` will not work as intended and the one not in a `gd` will work as intended. This is permanently buggy and also persistent in all screen sizes at this point until you refresh.

If you start at `lt-md` and start with a refreshed browser then everything works as intended and all is well. However once changing to `gt-sm`, chaos ensues if you try to return to `lt-md`

The Stack Blitz is broken right now but here is a Stack Blitz with a basic example of this:
[StackBlitz](https://stackblitz.com/edit/angular-flex-layout-seed-t8a7hz?file=app/app.component.ts)

Since the stack blitz is broken you can get the code from there and test it somewhere else, but here is a Pastebin with the code in case you wanted to test it. It's pretty simple so you can copy-paste in into any angular project that uses angular flex layout and it will work:
[PasteBin](https://pastebin.com/7BBBDgsa)

#### What is the use-case or motivation for changing an existing behavior?

Well, if `fxShow` and `fxHide` was intended to work with `gd` then this is just incorrect behavior.

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?

Angular 7 on Chrome is what I was using when I discovered this.

CaerusKaru:
No, grid directives are not as inter-connected with the rest of the library as the others. This was work that was planned, but postponed. We're working on some re-architecting right now (in the design phase still) that should correct that though. Stay tuned!

RyanTibbetts-OnScale:
[...]

Issue #1035, opened 2019-03-16
useColumnBasisZero:false not applied in production builds


gerardcarbo:
# Bug Report

#### What is the expected behavior?

When FlexLayoutModule.withConfig({useColumnBasisZero: false}) is specified, fxFlex default output shoud be flex:1 1 auto

#### What is the current behavior?

In production builds it still produces flex: 1 1 1e-9px

#### What are the steps to reproduce?

Install:

https://github.com/gerardcarbo/flex-layout-test

and run: 

`npm build && npm start`

The span where fxFlex is applied has the style: 

`style="flex: 1 1 1e-09px; box-sizing: border-box"`

If you run:

 `ng serve ` 

the same span has the style:

`style="flex: 1 1 auto; box-sizing: border-box;"`

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?

 ```
 "dependencies": {
    "@angular/animations": "~7.2.0",
    "@angular/cdk": "^7.3.4",
    "@angular/common": "~7.2.0",
    "@angular/compiler": "~7.2.0",
    "@angular/core": "~7.2.0",
    "@angular/flex-layout": "^7.0.0-beta.23",
    "@angular/forms": "~7.2.0",
    "@angular/platform-browser": "~7.2.0",
    "@angular/platform-browser-dynamic": "~7.2.0",
    "@angular/router": "~7.2.0",
    "core-js": "^2.5.4",
    "http-server": "^0.11.1",
    "rxjs": "~6.3.3",
    "tslib": "^1.9.0",
    "zone.js": "~0.8.26"
  },
  "devDependencies": {
    "@angular-devkit/build-angular": "~0.13.0",
    "@angular/cli": "~7.3.0",
    "@angular/compiler-cli": "~7.2.0",
    "@angular/language-service": "~7.2.0",
    "@types/node": "~8.9.4",
    "@types/jasmine": "~2.8.8",
    "@types/jasminewd2": "~2.0.3",
    "codelyzer": "~4.5.0",
    "jasmine-core": "~2.99.1",
    "jasmine-spec-reporter": "~4.2.1",
    "karma": "~3.1.1",
    "karma-chrome-launcher": "~2.2.0",
    "karma-coverage-istanbul-reporter": "~2.0.1",
    "karma-jasmine": "~1.1.2",
    "karma-jasmine-html-reporter": "^0.2.2",
    "protractor": "~5.4.0",
    "ts-node": "~7.0.0",
[...]

PR #1185, opened 2020-02-02
rfc: new layout system

CaerusKaru:


CaerusKaru:
Ref: #1141, #1130, #1161, #1035, #501

erkanmaras:
Is there any estimated release date for this ?

Splaktar:
> Is there any estimated release date for this ?

I don't believe so.

Splaktar:
I love the 70% bundle size reduction! I'm curious how much of that is JS vs CSS?

Would this be valid?
```html
  <bp name="gt-xs" layout="row" gap="25px">
```
Or do you need to do this?
```html
  <bp name="gt-xs" layout="row">
  <bp name="gt-xs" gap="25px">
```

1. What about `fxLayoutAlign`?
1. What about `fxFlex="33%"` and the other variations?
1. How does CSS Grid play into this?
  1. Would the library have an API that doesn't map directly to Grid/Flexbox and be able to determine when it was best to use Grid vs Flexbox?

CaerusKaru:
@Splaktar `<bp name="gt-xs" layout="row" gap="25px">` is totally valid, and is the intention of the design (consolidating by breakpoint).

* (and 2) What's special about `fxLayoutAlign` and `fxFlex="33%"`? They should translate to `align` and `flex` respectively. (`fxFlexAlign` translates to `flex-align`, but this is not dogmatic)
* CSS Grid is the exact same interface (e.g. alignColumns, gridGap, etc)
* No but no reason that can't be a follow-up feature

Splaktar:
Thank you for the quick response! I'm just starting to dig into the full `README.md`

src/lib/uni/README.md
```diff
@@ -0,0 +1,55 @@
...
+  imports: [UnifiedModule.withDefaults()] // brings in default tags and breakpoints
+})
+export class AppModule {}
+```
+
+### Step 2: Use the new syntax
+```html
+<div ngl flexAlign="start">
+  <bp tag="xs" flexAlign="end"></bp>
+  <bp tag="md" flexAlign="center"></bp>
```
Splaktar, line 27: The PR description uses `name="xs"` here. It looks like the rest of the code uses `tag="xs"`.

I'm not sure which one is clearer, probably `name`?

I guess the formal name for "the stuff inside the media query CSS at-rule" is a "media query list". But I don't think that we support multiple queries per `<bp>`. Perhaps it should just be `"query="xs"`?

src/lib/uni/README.md
```diff
@@ -0,0 +1,55 @@
...
+</div>
+```
+
+Here, `ngl` stands for "Angular (ng) Layout (l)". The `bp` elements are optional if
```
Splaktar, line 31: Was `ngl` selected just because it's shorter than `ngLayout`?

src/lib/uni/README.md
```diff
@@ -0,0 +1,55 @@
...
+you don't plan to use breakpoints in your template. Conceptually, it's cleaner than
+cramming all of the breakpoints on one root element, and the `bp` elements don't
```
Splaktar, line 33: `bp` is short, but I think that it lacks a little bit of expressiveness. It won't be readable to people who aren't familiar with this library's APIs.

Is `<breakpoint` overly verbose? It seems a lot more readable and expressive.


```html
<div ngLayout flexAlign="start">
  <breakpoint tag="xs" flexAlign="end"></breakpoint>
  <breakpoint tag="md" flexAlign="center"></breakpoint>
</div>
```

src/lib/uni/src/unified.ts
```diff
@@ -0,0 +1,200 @@
...
+    this.element = elementRef.nativeElement;
+  }
+}
+
+/**
+ * One directive to rule them all. This directive is responsible for
+ * tagging an HTML element as part of the layout system, and then
+ * coordinating all updates with GrandCentral.
+ */
+@Directive({selector: `[ngl]`})
```
Splaktar, line 48: This seems like a [structural directive](https://angular.io/guide/structural-directives) which should have a `*` in front of it (when used in a template), since it can change the structure of the DOM (especially when a directive like `order="2"` is used).

But then that introduces the limitation of [one structural directive per host element](https://angular.io/guide/structural-directives#one-structural-directive-per-host-element).

Have you considered this approach? Were issues uncovered?

src/lib/uni/src/central.ts
```diff
@@ -0,0 +1,175 @@
...
+/**
+ * GrandCentral is a switchyard for all of the various Layout directives
+ * registered in an application. It is the single source of truth for all of
+ * the layout changes that occur in an application. For instance, any changes
+ * to the browser state via registered media queries are monitored and updated
+ * in this service. The directives themselves simply store their respective values
+ * for each of the media states.
+ */
+@Injectable({providedIn: 'root'})
+export class GrandCentral {
```
Splaktar, line 28: `GrandCentral` is semi-obvious for Americans who live on the East Coast, but I'm not sure it's as universally well-known as we might want. Would something like `Orchestrator` or `Conductor` (I think that this could have minor collisions with other meanings of the word) be more widely understood?

src/lib/uni/src/breakpoint.ts
```diff
@@ -0,0 +1,145 @@
...
+ * corresponding mediaQuery, and a priority in case the mediaQuery overlaps
+ * with other registered breakpoints.
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/CSS/Media_Queries/Using_media_queries
+ */
+export interface Breakpoint {
+  /** The shortcut name for the breakpoint, e.g. 'xs' */
+  name: string;
+  /** The mediaQuery for the breakpoint, e.g. 'screen and (max-width: 500px)' */
+  media: string;
```
Splaktar, line 24: I think that there's a good case for calling this `query` or `mediaQuery`.

src/lib/uni/src/breakpoint.ts
```diff
@@ -0,0 +1,145 @@
...
+  provide: BREAKPOINTS,
+  useValue: DEFAULT_BREAKPOINTS,
+  multi: true,
+};
+
+/**
+ * An internal-facing injection token to consolidate all registered
+ * breakpoints for use in the application.
+ */
+export const BPS = new InjectionToken<Breakpoint[]>('Angular Layout Condensed Breakpoints', {
```
Splaktar, line 131: It's a short token, but it's not very expressive.

src/lib/uni/src/tags/tag.ts
```diff
@@ -0,0 +1,68 @@
...
+  priority: number;
+}
+
+/**
+ * A tag is a way of consolidating logic about a style pattern. For instance,
+ * setting the 'flex' attribute could be done with a Flex Tag. Each tag has an
+ * associated name, builder, cache, and dependencies on other builder input
+ * values.
+ */
+export abstract class Tag {
```
Splaktar, line 29: In this implementation, `Tag`s are attributes? I think that this can be easily confused with [HTML tags](https://developer.mozilla.org/en-US/docs/Glossary/Tag).

samal-rasmussen:
Does this implementation reduce the amount of layout thrashing compared to what the existing implementation does?

The performance issues with the use of inline styles in the existing implementation is discussed here: https://github.com/angular/flex-layout/issues/1299

fxck:
Please do yourself a favor and get rid of flex-layout as soon as possible, it HUGELY affects performance, especially when nested inside other components. Imagine 10 cards in a grid layout, with 3 components inside, each having few components with another grid inside. It's not unreasonable amount of components and grid layout required along with it, even with just a little more complex UI elements. If you rely on flex-layout for this, you **are** going to make your app unusable. 

We realized this way too late and the refactoring was costly. Both in time and mental health. Get out while you can.

image

CaerusKaru:
@fxck Please consider moving your comment to the discussion thread, since your comment has no bearing on this PR.

@samal-rasmussen Last I remember, this does not incorporate a class-based approach for a few reasons. 1) was to semi-maintain parity with the existing toolset, which prioritizes style CSS rank vs performance, and 2) there are certain features that simply can't be handled with a class-based approach. This does not mean we won't add that functionality later, or in the current version of the library; it's distinct from this proposal, which is more about how to structure the directives themselves in the markup.

epelc:
@CaerusKaru are there any plans for an automatic upgrade using `ng update` or similar? The feature set seems largely the same and like this could be supported.

Also is there a link to the discussion thread? I could find anything here.

CaerusKaru:
@epelc I would love to have some `ng update` functionality, and if we ever invested in this actively, that would definitely be a fast follow to release (again, if it were up to me). Feel free to add discussion directly to this PR. It would be too tedious to go back and forth on certain points between here and a tracking issue.

ckorherr:
> @Splaktar `<bp name="gt-xs" layout="row" gap="25px">` is totally valid, and is the intention of the design (consolidating by breakpoint).
> 
> * (1 and 2) What's special about `fxLayoutAlign` and `fxFlex="33%"`? They should translate to `align` and `flex` respectively. (`fxFlexAlign` translates to `flex-align`, but this is not dogmatic)
> * CSS Grid is the exact same interface (e.g. alignColumns, gridGap, etc)
> * No but no reason that can't be a follow-up feature

One thing is special about the `align` attribute, it is really badly rendered in vscode, see: https://github.com/Microsoft/vscode/issues/66723

PR #1226, opened 2020-03-31
build: do not include secondary entry-points in primary UMD bundle

devversion:
The APF specification does not capture all corner cases, but conceptually,
generated UMD bundles should match the actual entry-point in the typings tree.
This means that it's technically spec-compliant to include secondary entry-points in a
primary UMD bundle. At the same time though, it's also spec-compliant to not include
secondary entry-points and require UMD consumers to load the secondary entry-point
UMD bundles manually. 

We are going with the latter since that helps with an issue that has surfaced in ngcc.
The plan is to update the APF specification to explicitly guard against cases like that,
so that primary UMD bundles no longer can bundle secondary entry-points. This
change prepares flex-layout for the change to the APF specification.

### DIFF
--- a/tools/package-tools/build-bundles.ts
+++ b/tools/package-tools/build-bundles.ts
@@ -9,7 +9,6 @@ import {remapSourcemap} from './sourcemap-remap';
 // There are no type definitions available for these imports.
 const rollup = require('rollup');
 const rollupNodeResolutionPlugin = require('rollup-plugin-node-resolve');
-const rollupAlias = require('rollup-plugin-alias');
 
 /** Directory where all bundles will be created in. */
 const bundlesDir = join(buildConfig.outputDir, 'bundles');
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@devversion, @CaerusKaru 
From a developer perspective, what is the impact of this change? 
What do they need to do [differently] at the application level in their projects?

devversion:
@ThomasBurleson The majority of developers is not affected by this change. Angular CLI projects by default will always use the flat `esm2015` output. This PR only only changes the UMD bundles, and generally UMD bundles are not used, unless someone builds an Angular project without bundling.

In those cases, developers need to load a few more UMD bundles (for the secondary entry-points):

1. `@angular/flex-layout/bundles/flex-layout-core.umd.js`.
2. `@angular/flex-layout/bundles/flex-layout-flex.umd.js`.
3. `@angular/flex-layout/bundles/flex-layout-grid.umd.js`.
4. `@angular/flex-layout/bundles/flex-layout-extended.umd.js`

This just highlights the problem with UMD though. Ideally these projects would either use fine-grained entry-points, or not use UMD at all. e.g. only using `@angular/flex-layout/grid`.

I'm not too happy with the developer experience on UMD bundles, but we had to go down this route as re-exporting secondary entry-points seems to be rarely used in libraries, and NGCC does not support this.

PR #1236, opened 2020-05-07
fix(match-media): unregister media query event listeners on destroy

### LINKED ISSUE #1213: Memory leak with media query listeners
# Bug Report

Registered event handlers are unregistered on destroy to be able to free memory.

This media query listener is never unregistered, leasing to memory leaks:

https://github.com/angular/flex-layout/blob/36f8f390acbb38bc78a7d762d23799b32a47b0c9/src/lib/core/match-media/match-media.ts#L115

In unit tests it's typical to setup and teardown an application, in which case memory from a previous run should not be retained.

Observed in flex-layout@9.0.0-beta.29 with Angular 9.0.4.

CaerusKaru:
When the internal registry of media query listeners is reset, or
MatchMedia is destroyed, the listeners themselves are not currently
being removed. This leads to memory leaks, which is corrected by
this commit.

Fixes #1213

### DIFF
--- a/src/lib/core/match-media/match-media.ts
+++ b/src/lib/core/match-media/match-media.ts
@@ -5,7 +5,7 @@
  * Use of this source code is governed by an MIT-style license that can be
  * found in the LICENSE file at https://angular.io/license
  */
-import {Inject, Injectable, NgZone, PLATFORM_ID} from '@angular/core';
+import {Inject, Injectable, NgZone, OnDestroy, PLATFORM_ID} from '@angular/core';
 import {DOCUMENT, isPlatformBrowser} from '@angular/common';
 import {BehaviorSubject, Observable, merge, Observer} from 'rxjs';
 import {filter} from 'rxjs/operators';
@@ -20,10 +20,11 @@ import {MediaChange} from '../media-change';
  * NOTE: both mediaQuery activations and de-activations are announced in notifications
  */
 @Injectable({providedIn: 'root'})
-export class MatchMedia {
+export class MatchMedia implements OnDestroy {
   /** Initialize source with 'all' so all non-responsive APIs trigger style updates */
   readonly source = new BehaviorSubject<MediaChange>(new MediaChange(true));
   registry = new Map<string, MediaQueryList>();
+  private readonly pendingRemoveListenerFns: Array<() => void> = [];
 
   constructor(protected _zone: NgZone,
               @Inject(PLATFORM_ID) protected _platformId: Object,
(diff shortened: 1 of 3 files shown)


src/lib/core/match-media/match-media.ts
```diff
@@ -43,6 +43,15 @@ export class MatchMedia {
     return results;
   }
 
+  get registry(): Map<string, MediaQueryList> {
+    return this._registry;
+  }
+
+  set registry(registry: Map<string, MediaQueryList>) {
+    this.emptyRegistry();
```
JoostK, line 51: I think this is problematic. Consider the scenario where `MediaTrigger` "steals" ownership of the registry to force different activations. This would empty the registry here and unregister the listener functions, however when the `MediaTrigger` then restores the registry using the media queries it set aside, those will effectively be "dead" as no more listeners are registered.

Either `MediaTrigger` needs to take over full ownership and be made responsible for its own cleanup, or cleanup needs to be managed separately from activations. I think the latter is the simplest approach, with the catch that "borrowed" media queries in `MediaTrigger` will be unregistered when destroying `MatchMedia`. I think this is fine though, as `MediaTrigger` itself doesn't require the `MediaQueryList` to have a listener, *and* the listeners were registered in `MatchMedia` in the first place.

I prototyped this in https://github.com/JoostK/flex-layout/commit/38e26679387038e76901214949878d764ea4e1fd, which is a fixup commit of this PR's commit. With that approach, the getter/setter dance of `this.registry` is no longer strictly required, nor are the changes to `MediaTrigger.restoreRegistryMatches`.

src/lib/core/media-trigger/media-trigger.ts
```diff
@@ -174,13 +172,14 @@ export class MediaTrigger {
    * Restore original, 'true' registry
    */
   private restoreRegistryMatches() {
-    const target = this.matchMedia.registry;
+    const target = new Map(this.matchMedia.registry);
```
JoostK, line 175: This change is no longer necessary. In fact, the below `target.clear()` makes cloning the registry rather unnecessary.

Either these changes can be reverted in full, or the clone + clear can be refactored. I would personally revert these changes.

src/lib/core/media-trigger/media-trigger.ts
```diff
@@ -151,7 +149,7 @@ export class MediaTrigger {
   private forceRegistryMatches(queries: string[], matches: boolean) {
     const registry = new Map<string, MediaQueryList>();
     queries.forEach(query => {
-      registry.set(query, {matches: matches} as MediaQueryList);
+      registry.set(query, {matches: matches, removeListener: (_) => {}} as MediaQueryList);
```
JoostK, line 152: This shouldn't be necessary anymore, given that removal functions are only registered when creating a real media query list.

PR #1252, opened 2020-05-13
fix(show-hide): set explicit display fallback for SSR

### LINKED ISSUE #1250: Show/Hide bug beta30  caused by display:initial
# Bug Report
In beta30 the following commit causes issues with show/hide because of `display:initial`

>  show-hide: use initial value as fallback instead of parent (#1243) (bf2355b) 

the footer should be centered and displaying as it was with beta29

See stackblitz

see the following stackblitz to reproduce. I hide the footer by default and `show.gt-sm`. 

https://stackblitz.com/edit/angular-flex-layout-seed-jyuaqd

Using the latest versions

CaerusKaru:
In the event that a user specifies two breakpoints with conflicting
show/hide, e.g. fxHide.lt-lg and fxShow.md, fxShow will not be
able to re-enable CSS `display`, since there's no equivalent to
nulling out the styles, which is what we do when handling it
inline in the browser.

This sets the property to `initial` in SSR stylesheets only. This
should preserve browser behavior where nulling out the styles still
works.

NOTE: if you have other classes that set the `display` property of
an element, they may get overridden in SSR only. This is
unavoidable, since getComputedStyle is not available to use in SSR.

Fixes #1250

### DIFF
--- a/src/lib/extended/show-hide/show-hide.ts
+++ b/src/lib/extended/show-hide/show-hide.ts
@@ -30,13 +30,14 @@ import {takeUntil} from 'rxjs/operators';
 
 export interface ShowHideParent {
   display: string;
+  isServer: boolean;
 }
 
 @Injectable({providedIn: 'root'})
 export class ShowHideStyleBuilder extends StyleBuilder {
   buildStyles(show: string, parent: ShowHideParent) {
     const shouldShow = show === 'true';
-    return {'display': shouldShow ? parent.display : 'none'};
+    return {'display': shouldShow ? parent.display || (parent.isServer ? 'initial' : '') : 'none'};
   }
 }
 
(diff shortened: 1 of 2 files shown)


sjbeatle:
@CaerusKaru 
Hey! Given your note, do you have any suggestions on what to do?

For example:
```
// Template
<div class="make-me-display-flex" fxHide.xs>Hello World!</div>
// CSS
.make-me-display-flex { display: flex; }
```
This change now injects and renders the following inline style for all shown breakpoints:
```
// Rendered HTML
<div class="make-me-display-flex" style="display: initial;">Hello World!</div>
```
Which kills the CSS display flex :-(

Current workarounds (neither ideal):
 A. Wrap the div and apply the directive there  <-- preferable of the two
```
// Template
<div fxHide.xs>
  <div class="make-me-display-flex">Hello World!</div>
</div>
```
 B. If you preempt with your own inline style, flex-layout will recognize and use that <-- double-duty, yuck!
```
// Template
<div class="make-me-display-flex" style="display: flex;" fxHide.xs>Hello World!</div>
```

Any advice would be greatly appreciated. Thanks!

CaerusKaru:
@sjbeatle We've published 9.0.0-beta.31 to revert the change and add this fix.

PR #1261, opened 2020-05-18
build: upgrade to Angular v10, CDK v10, and TypeScript 3.9

CaerusKaru:
This also adds tslib as a direct dependency, since TypeScript 3.9
is incompatible with older versions of tslib.


devversion:
@googlebot I consent.

PR #1308, opened 2020-11-12
fix(core): set fractional breakpoints to .98 instead of .9

CaerusKaru:
Setting the breakpoint value to .9 leads the browser to incorrectly
round-up and cause overlap with fractional pixels and their
upper bound. This leads to unintended consequences in downstream
consumers like our MediaObserver and MatchMedia providers.


CaerusKaru:
Ref: https://github.com/angular/components/issues/20850

PR #1353, opened 2021-07-13
build: export grid directives

### LINKED ISSUE #1352: Grid Public Api
# Bug Report

Expected behaviour is for grid directives to be exposed from public api of library

Grid directives are not exposed from public api of library

[StackBlitz](https://stackblitz.com/edit/angular-ivy-vkpeb7?file=src/app/custom-grid.directive.ts)

When I tried to use custom breakpoints with grid directives, I revealed a bug with imports, I couldn't import directives because they are not exposed from public api of library. So I added needed imports in this [PR](https://github.com/angular/flex-layout/pull/1353).

ruslic19:
Fixes #1352

### DIFF
--- a/src/lib/grid/public-api.ts
+++ b/src/lib/grid/public-api.ts
@@ -8,4 +8,14 @@
 
 export * from './module';
 
-
+export * from './align-columns/align-columns'
+export * from './align-rows/align-rows'
+export * from './area/area'
+export * from './areas/areas'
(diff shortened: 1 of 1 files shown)


ruslic19:
@googlebot I signed it!

CaerusKaru (approved):

CaerusKaru:
@ruslic19 Thanks for this contribution!

PR #1369, opened 2021-11-07
build: upgrade to Angular v13 and APF v13, emit Ivy artifacts

CaerusKaru:


tools/package-tools/gulp/build-scss-task.ts
```diff
@@ -2,7 +2,7 @@ import {src, dest} from 'gulp';
 import {join} from 'path';
 
 // These imports lack of type definitions.
-const gulpSass = require('gulp-sass');
+const gulpSass = require('gulp-sass')(require('node-sass'));
```
devversion, line 5: Node-sass 😱 😄

tools/package-tools/build-package.ts
```diff
@@ -81,19 +81,18 @@ export class BuildPackage {
...
   async createBundles() {
     await this.bundler.createBundles();
   }
 
-  /** Compiles TS into both ES2015 and ES5, then updates exports. */
+  /** Compiles TS into both ES2015 and ES2020, then updates exports. */
   private async _compileBothTargets(p = '') {
     return compileEntryPoint(this, buildTsconfigName, p)
-        .then(() => compileEntryPoint(this, buildTsconfigName, p, this.esm5OutputDir));
+        .then(() => compileEntryPoint(this, buildTsconfigName, p, this.esm2020OutputDir));
```
devversion, line 95: Should that be `esm2015OutputDir` according to `compileEntryPoint`? I think ES2020 should be default, and then the sub-folder /second compilation is es2015.

devversion (commented):
For CI failures, it might be because of using an older Yarn version?

devversion (approved):

PR #1371, opened 2021-11-18
build: migrate to use Angular CLI

alan-agius4:


CaerusKaru:
Merged #1372, with full co-author credit to @alan-agius4

Issue #1396, opened 2022-03-01
Can't find stylesheet to import error for @angular/flex-layout/mq


DenysVuika:
# Bug Report

#### What is the expected behavior?

Library should compile without errors

#### What is the current behavior?

Angular 13 library fails to compile with `@angular/flex-layout@13.0.0-beta.38` 

```
✖ Compiling with Angular sources in Ivy full compilation mode.
Can't find stylesheet to import.

 @import '@angular/flex-layout/mq';
```

#### What are the steps to reproduce?
Providing a StackBlitz (or similar) is the *best* way to get the team to see your issue.

*  [StackBlitz starter](https://stackblitz.com/edit/flex-layout-seed-nightly) using latest `npm` nightly release: 

#### Is there anything else we should know?

Project compiles fine with 12.x version of the flex-layout

juane1000:
I am also having this issue

burnaDLX:
I changed the path from `@angular/flex-layout/mq` to `node_modules/@angular/flex-layout/core/sass/layout-bp` and it seems to work.

CaerusKaru:
Can someone provide a minimal reproduction of the issue? If possible, I'd like to include a fix for this in the v14 release.

Issue #1394, opened 2022-02-22
fxLayout=column get rendered as row in Angular 13 Universal


vnistorica:
# Bug Report

#### What is the expected behavior?
The result from universal of fxLayout="column", should be displaying the containers one under the other.

#### What is the current behavior?
They are displayed as row layout initially. After the javascript gets loaded the layout returns to the correct layout.

![Screenshot from 2022-02-22 15-21-59](https://user-images.githubusercontent.com/100215532/155153091-fc5f98cb-8bf2-4fac-8f24-dc9d040fe884.png)

#### What are the steps to reproduce?
Add the following code, having Angular 13.2.2 and flex layout 13.0.0-beta.38.
```
<div fxLayoutGap="10px" fxLayout="column" fxFlexFill fxLayoutAlign="start stretch">
    <div style="padding: 100px; background-color: #00ccbc">Test</div>
    <div style="padding: 100px; background-color: #6d2275">Test</div>
    <div style="padding: 100px; background-color: #4f694c">Test</div>
    <div style="padding: 100px; background-color: #bb8d8d">Test</div>
</div>
```
Disable browser javascript to get the result of SSR.

#### Which versions of Angular, Material, OS, TypeScript, browsers are affected?
Node: v16.13.2
Npm: 8.1.2
OS: linux x64

```
 "dependencies": {
    "@angular-material-extensions/password-strength": "~10.1.0",
    "@angular/animations": "~13.2.2",
    "@angular/cdk": "~13.2.2",
    "@angular/common": "~13.2.2",
    "@angular/compiler": "~13.2.2",
    "@angular/core": "~13.2.2",
    "@angular/flex-layout": "~13.0.0-beta.38",
    "@angular/forms": "~13.2.2",
    "@angular/localize": "~13.2.2",
    "@angular/material": "~13.2.2",
    "@angular/platform-browser": "~13.2.2",
    "@angular/platform-browser-dynamic": "~13.2.2",
    "@angular/platform-server": "~13.2.2",
    "@angular/router": "~13.2.2",
    "@nguniversal/express-engine": "^13.0.1",
    "@ngx-translate/core": "^13.0.0",
    "@ngx-translate/http-loader": "^6.0.0",
    "compression": "^1.7.4",
    "core-js": "^3.6.4",
    "express": "^4.15.2",
    "form-data": "^4.0.0",
    "luxon": "^2.2.0",
    "morgan": "^1.10.0",
    "ngx-cookie-service": "^13.0.0",
[...]

PR #1405, opened 2022-06-03
chore: upgrade to Angular 14.0.0

SamuelMarks:
Issues:
> DEPRECATED: The 'defaultProject' workspace option has been deprecated. The project to use will be determined from the current working directory.

…and:

> Package "@nguniversal/builders" has an incompatible peer dependency to "@angular-devkit/build-angular" (requires "^13.0.2", would install "14.0.0").
                  Package "@ngtools/webpack" has an incompatible peer dependency to "typescript" (requires "~4.4.3", would install "4.7.2").

---

Approach:

```sh
$ git clone --depth=1 https://github.com/angular/flex-layout
$ git remote add mine https://github.com/SamuelMarks/flex-layout
$ yarn
$ ng update --force $(jq -r '.dependencies * .devDependencies | keys | map(select(startswith("@angular"))) | @tsv' package.json)
$ git checkout -b ng14
$ git commit -S -am 'chore: upgrade to Angular 14.0.0'
$ git push --set-upstream mine ng14
```


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "flex-layout-srcs",
   "description": "Angular Flex Layout",
-  "version": "13.0.0-beta.38",
+  "version": "14.0.0-beta.38",
```
michaelfaith, line 4: They usually increment the beta version on new releases.  
```suggestion
  "version": "14.0.0-beta.39",
```

PzYon:
Hi there. Any idea when we can expect the release of a new version? Would be really great so we can update to Angular 14. Thank you very much.

CaerusKaru:
There are a couple of issues that I'd like to resolve before we cut a release. I have no issue if @SamuelMarks would like to remove the version bump for this, that way at least the nightly builds will support v14. But I want to close out at least #1396 and #1394, which were regressions of the last version. I've slated some time for this week to investigate these, and if all goes to plan, we could see a release sometime next week. In the meantime, there are some NPM and Yarn workarounds you can use to migrate.

Also big thanks to @SamuelMarks for taking on this work (this PR, not the issues above), it is much appreciated.

damingerdai:
what time will the pr be merged?

kuncevic:
@CaerusKaru Running `yarn` then after it fail I am running `yarn run postinstall --force` which is fixing local build for now
However this workaround does not work for github actions, any ideas how to please gihub actions?

CaerusKaru:
Thinking the browserstack issues might be caused by permissions of some sort, I opened #1409 (with some other stuff I wanted to include), but that's failing too with the same issue. I'll reach out to the dev infra team to help resolve this.

michaelfaith:
@CaerusKaru for this and future updates, would it be possible to use peer deps similar to how Angular Material does, which includes the next major version as well, to easy the update process.  e.g. `^14.0.0 || ^15.0.0` (https://github.com/angular/components/blob/5816a12dfa816e3f340627fdf561658de0e8c970/packages.bzl#L4)  It's not usually the case that the next major Angular or CDK versions break this library, so it's likely safe to do and would improve the DX for this interim period?

JanneHarju:
I haven't use CircleCI so not sure how it should work but there is two Errors even there is Success in the end. Both errors say this "DEPRECATION: Passing custom equality testers to MatchersUtil#equals is deprecated." I point it out here if you missed it. https://app.circleci.com/pipelines/github/angular/flex-layout/270/workflows/10db3ad6-18ef-47d8-8dee-9c25b572db2a/jobs/7792

projects/libs/flex-layout/package.json
```diff
@@ -29,7 +29,7 @@
     "tslib": "^2.3.0"
   },
   "peerDependencies": {
-    "@angular/cdk": "^13.0.0",
+    "@angular/cdk": "^14.0.0",
```
michaelfaith, line 32: Would it be possible to use peer deps similar to how Angular Material does, which includes the next major version as well, to ease the update process.  e.g. `^14.0.0 || ^15.0.0`.  This should be safe and would improve the DX for this interim period between angular and flex layout releases.

kasir-barati:
any progress on this? 

```cmd
pnpm add @angular/flex-layout @angular/cdk
Already up-to-date
Progress: resolved 543, reused 523, downloaded 0, added 0, done
 WARN  Issues with peer dependencies found
.
└─┬ @angular/flex-layout
  ├── ✕ unmet peer @angular/cdk@^13.0.0: found 14.0.2
  ├── ✕ unmet peer @angular/core@^13.0.0: found 14.0.1
  ├── ✕ unmet peer @angular/common@^13.0.0: found 14.0.1
  └── ✕ unmet peer @angular/platform-browser@^13.0.0: found 14.0.1
```

It is really annoying and unstable to use `--force`

projects/libs/flex-layout/package.json
michaelfaith, line 32: Pardon the repeated comment, I wanted to point out the specific line this was referring to

> @CaerusKaru for this and future updates, would it be possible to use peer deps similar to how Angular Material does, which includes the next major version as well, to ease the update process. e.g. `^14.0.0 || ^15.0.0` (https://github.com/angular/components/blob/5816a12dfa816e3f340627fdf561658de0e8c970/packages.bzl#L4) It's not usually the case that the next major Angular or CDK versions break this library, so it's likely safe to do and would improve the DX for this interim period between angular and flex layout releases.

projects/libs/flex-layout/package.json
```diff
@@ -29,7 +29,7 @@
     "tslib": "^2.3.0"
   },
   "peerDependencies": {
-    "@angular/cdk": "^14.0.0",
+    "@angular/cdk": "^13.0.0 || ^14.0.0",
```
michaelfaith, line 32: @SamuelMarks it should go the other way. ^14.0.0 || ^15.0.0 that way when Angular 15 comes out, people with Flex-Layout 14 won't have trouble upgrading to angular 15, while this team works on updating flex-layout to 15

SamuelMarks: Ah, ok just did `fastmod -F '^13.0.0 || ^14.0.0' '^14.0.0 || ^15.0.0'`

PR #1409, opened 2022-06-14
build: upgrade to Angular and CDK v14

CaerusKaru:
Also bump TS target to ES2020 and remove deprecated angular.json
option.


luiz-limber:
up

michaelfaith:
I made this comment in the other PR, but noticed this one was created to.  In case you didn't see it in the other one: for this and future updates, would it be possible to use peer deps similar to how Angular Material does, which includes the next major version as well, to ease the update process.  e.g. `^14.0.0 || ^15.0.0` (https://github.com/angular/components/blob/5816a12dfa816e3f340627fdf561658de0e8c970/packages.bzl#L4)  It's not usually the case that the next major Angular or CDK versions break this library, so it's likely safe to do and would improve the DX for this interim period between angular and flex layout releases.

I believe that that would mean updating `requiredAngularVersion` and `peerDependencies["@angular/cdk"` in the `package.json` to both be `^14.0.0 || ^15.0.0`  instead of just `^14.0.0`.

chamikasandamal:
Can someone please have a look on this check for PR

PzYon:
Looks like the `yarn.lock` file needs to be updated: `error Your lockfile needs to be updated, but yarn was run with --frozen-lockfile`. As this causes thie `lint`- and `build`-jobs to fail:

https://app.circleci.com/pipelines/github/angular/flex-layout/291/workflows/0086ee1c-fa87-43db-a469-4fd5edec7c66/jobs/7885

I guess updating the `yarn.lock` locally and then commiting it will help?

Sebastian-G:
@PzYon But e.g. in my case it wasn't an issue to build it according to the `yarn.lock`.
I used yarn v1.22.19 to install, instead of v1.22.17 🤔 

```
➜ yarn install --frozen-lockfile --non-interactive
yarn install v1.22.19
[1/5] 🔍  Validating package.json...
[2/5] 🔍  Resolving packages...
[3/5] 🚚  Fetching packages...
[4/5] 🔗  Linking dependencies...
[5/5] 🔨  Building fresh packages...
✨  Done in 155.94s.
```

CaerusKaru:
As an update to everyone following here, I have been trying and researching in vain to try and get to the bottom of the BrowserStack issues. Since SauceLabs (and native headless testing) gives us good coverage on their own, I've asked @josephperrott to disable the `Required` label for `tests_browserstack`, so that we can merge this PR. Until then, unfortunately, we are frozen (unless some other intrepid soul can suss out the reason for the failure). If this drags on too long, I'll simply remove the tests_browserstack logic so it becomes a passthrough, but I'd prefer to leave it in place so that we can still investigate asynchronously.

Sebastian-G:
@michaelfaith not sure if the 2 min count in the installation. 
The timer seems to start straight after `browserstack-tunnel/BrowserStackLocal` was started. 

See https://github.com/angular/flex-layout/blob/v14/scripts/browserstack/start-tunnel.sh

michaelfaith:
> @michaelfaith not sure if the 2 min count in the installation. The timer seems to start straight after `browserstack-tunnel/BrowserStackLocal` was started.
> 
> See https://github.com/angular/flex-layout/blob/v14/scripts/browserstack/start-tunnel.sh

@Sebastian-G Not sure about that.  If you hover over the colored bars it says what each bar represents.  I'm just going by that chart.

image

luiz-limber:
image

where is the error? 🤷🏻‍♂️🤷🏻‍♂️😭😭

CaerusKaru:
> image
> 
> where is the error? 🤷🏻‍♂️🤷🏻‍♂️😭😭

Story of my life 🤦

midgleyc:
Hi, just a post because I'm also interested in this getting merged.

Might the test script not finishing be because it's running in watch mode by default? I remember this being a problem a while back (e.g. https://github.com/angular/angular-cli/issues/978).

I think you can disable watch mode by running `ng test --no-watch`.

Failing that, you could set `singleRun: true` in `karma.conf.js`, but I don't know what else relies on that, or even really what it does.

CaerusKaru:
@josephperrott has disabled the required-ness of browserstack tests, so I'm going to close this PR and open a new one to pick up this change. Will post link when it's ready.

PR #1415, opened 2022-06-30
chore: update to 14.0.0-beta.39 with changelog

CaerusKaru:


luiz-limber:
thank you  guys
who will do the npm update? 😇

CaerusKaru:
Let's just say that I expect the number of possibly impacting breaking changes to increase in the next few versions, specifically targeting APIs that we use in this library (and that angular/components may not use). On top of this, the components team has a full-time staff that can adjust, discuss, and react quickly to breaking changes as they're being planned and rolled out. While I am also privy to the same discussion and planning, I am not a full time resource for this project, as much as I sometimes wish I could be. So while I understand your position as a longtime user, it's just not something I can realistically put on the library right now. If the Angular team changes their position on the support story for this repo, that's of course something we can re-examine.

